PR #1, opened 2021-03-18
Add python linter etc. files

pdiakumis:
Setting up Python linter/pre-commit files according to suggestions in team-docs.


lgruen (approved):
Nice!

PR #2, opened 2021-04-06
Add manual upload management scripts

lgruen:
Please let me know if you have any ideas regarding the folder structure in these buckets. E.g. where are we going to put the SNP data?

### DIFF
--- a/scripts/copy_test_gvcfs.py
+++ b/scripts/copy_test_gvcfs.py
@@ -7,5 +7,5 @@ output = os.getenv('OUTPUT')
 assert output and output.startswith('gs://cpg-tob-wgs-test/')
 
 subprocess.run(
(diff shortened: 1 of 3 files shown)


scripts/copy_test_gvcfs.py
```diff
@@ -0,0 +1,11 @@
+"""Copies a subset of gVCFs from the main bucket to the test bucket."""
```
pdiakumis, line 1: Nitpick: did we settle on gVCF or GVCF?

pdiakumis (approved):

scripts/copy_test_gvcfs.py
lgruen: We didn't -- the GATK documentation itself is totally inconsistent. See e.g. https://gatk.broadinstitute.org/hc/en-us/articles/360036827351-CombineGVCFs :)

PR #5, opened 2021-04-13
Move to batch-based directory structure

lgruen:


scripts/move_uploaded_crams.py
```diff
@@ -5,6 +5,11 @@
 import subprocess
 
 output = os.getenv('OUTPUT')
-assert output and output.startswith('gs://cpg-tob-wgs-archive/')
+assert output and output.startswith('gs://cpg-tob-wgs-archive/cram/batch')
```
vladsavelyev, line 8: Should it be `batch` or `batch0`?

lgruen: Good spotting :). This doesn't contain the `0` so we can run it for future batches too :)

vladsavelyev: Ah, gotcha. So in the command line, you would set it to `gs://cpg-tob-wgs-main/gvcf/batch0/`. Same for `move_uploaded_gvcfs.py`.

vladsavelyev (approved):

scripts/move_uploaded_crams.py
lgruen: That's exactly right.

PR #6, opened 2021-04-19
Copy test VCFs: hardcode output destination

vladsavelyev:
Would it make sense to copy more data into the test bucket, to properly test RF and VQSR?


lgruen:
> Would it make sense to copy more data into the test bucket, to properly test RF and VQSR?

To properly test the results of RF and VQSR, I understand that you need more samples (probably even more than we have in total now). That wasn't really the intention of the `test` access level / bucket though -- it's really meant for making sure your code runs properly and you're not slowed down by code reviews before submitting a full run on the main bucket.

Could you instead run your code with `standard` permissions? You could write to the `temporary` bucket if you don't expect the results to be useful longer-term.

vladsavelyev:
Fair!

Still, the script needs an update to reflect the new folder structure. Pushed another commit.

vladsavelyev:
Ah, just realized I didn't have the https://github.com/populationgenomics/tob-wgs/pull/5 pulled locally, and the script already reflects the new project structure. 

Anyway, there is a `check=False` to fix linting.

lgruen (approved):
Maybe change the PR title, but then good to go!

PR #7, opened 2021-04-23
Liftover SNPchip VCF to GRCh38

pdiakumis:
Adding an exploratory report to test out concordance for three samples as the title suggests. Rendered report is at: http://cpg.pdiakumis.com/tobwgs/concordance/

I'll try and set up a separate branch to hold the HTML so we can host these on GitHub using GitHub Pages.


lgruen:
> I'll try and set up a separate branch to hold the HTML so we can host these on GitHub using GitHub Pages.

Could we potentially use `nbviewer`?
https://docs.github.com/en/github/managing-files-in-a-repository/working-with-jupyter-notebook-files-on-github

scripts/snpchip_concordance.py
```diff
@@ -0,0 +1,43 @@
...
+
+mt = hl.import_vcf(INPUT_VCF)
+
+# Explore MT
+mt.count()
+mt.rows().show()
+mt.s.show(5)
+mt.entry.show(5)
+
+
```
lgruen, line 22: Debug remnants?

```suggestion
```

lgruen (commented):
Would probably be good to get this reviewed by someone who knows R better than me :).

How will we render the Rmd into an HTML file in the `web` bucket? (Would be good to update your PR description accordingly.)

scripts/snpchip_concordance.py
pdiakumis: No, not debugging, just exploring. I presume those will show up in the Batch logs. Or am I doing something wrong? Input is in the analysis bucket.

pdiakumis:
> Would probably be good to get this reviewed by someone who knows R better than me :).
> 
> How will we render the Rmd into an HTML file in the `web` bucket? (Would be good to update your PR description accordingly.)

Sorry, this is all part of a larger PR. Main focus of this was the exploration of the lifted over VCF. But not sure how to organise these files when they're all part of the same analysis.

pdiakumis:
Removed those extra files, will put in separate PRs.

scripts/snpchip_concordance.py
lgruen: Exploring data this way is not ideal because of the PR friction that's involved for accessing data in the `analysis` bucket. I think a better process would be to create a subset of this data that's stored in the `test` bucket and use a notebook to perform this exploration. I.e. whatever created this result in the `analysis` bucket should ideally also have created a smaller version of that dataset for the `test` bucket. One way to do this is to parametrize that script on the `ACCESS_LEVEL` environment variable and either read from `main` or `test` / write to `analysis` or `test` accordingly. Would that work?

pdiakumis: Sorry, I'm just confused with the `analysis` bucket. When should we use it? Or is it redundant now?
I'm also confused with how the `main` subsetting works. Who decides what gets filtered out/which subset to use?

lgruen: The idea behind the distinction of the `main` vs `analysis` bucket is in the way the `standard` access level is configured: you can _read_ from `main`, but you can _write_ to `analysis`. You currently can only _write_ to `main` when using the `full` access level. Maybe that additional safeguard isn't worth the complication that it creates though, especially since that setup isn't fully mirrored for `test` currently.

Whoever creates a new dataset in `main` / `analysis` should ideally also consider whether that data will be used for further downstream analysis. If so, having a subset in `test` seems handy. It's at that point where one needs to decide what subset makes sense for `test` -- it should be as small as possible, while still being useful for prototyping.

One way this can happen "naturally" is to either read data from `test` or from `main`, depending on the `ACCESS_LEVEL`. If e.g. an input matrix table then consists of only 20 samples in `test`, the derived dataset will automatically be subset accordingly.

PR #9, opened 2021-05-05
Update metadata files for batch 0 and 1

lgruen:
One of the metadata columns was swapped. Shyam uploaded the fixed versions.

@vladsaveliev, do we need to regenerate the MTs?


vladsavelyev (approved):
To summarise, it's:

1. `sample.flowcell_lane` mixed with `sample.platform`
2. `sample.id` values changed
3. `raw_data.PlinkSex` removed

None of those we used in the pipeline, so safe to replace.

lgruen:
> None of those we used in the pipeline, so safe to replace.

Thanks for verifying! 👍

lgruen:
FYI: This has been run.

PR #10, opened 2021-05-10
Copy SNPchip data to main bucket

pdiakumis:


lgruen (approved):
Thanks! Feel free to use bash scripts now as well :).

Please don't forget to delete your copy of the data after running this.

PR #11, opened 2021-05-10
add shebang

pdiakumis:


illusional (changes requested):
Might want to specify a specific Python, eg 3 or 2

illusional (approved):

vladsavelyev:
We probably shouldn't be supporting python2 at all

PR #16, opened 2021-05-23
Run the QC Rmd report with the anslysis-runner

vladsavelyev:
* Pulls metadata from the analysis bucket for now, which I manually populated. Until we have a proper metadata source.
* Puts the report into the web bucket (gotta need to version it and connect it with your web page).
* I removed the IS stddev plot for now, cause we didn't have that `InsertSizeStandardDeviationBatch1-2` file.


pdiakumis (approved):
Thank you, looks awesome!

PR #24, opened 2021-05-27
Move GATK4 pipeline validation results from the upload bucket

lgruen:


illusional (approved):
Cool use of `$OUTPUT`!

PR #30, opened 2021-05-28
Colour sina points

pdiakumis:
- use aesthetics for sina colours and shape
- refactor code


pdiakumis:
@vladsaveliev try it out if you want and we can customise colours/shapes/sizes. Not sure if we should put low quality samples in a separate table underneath instead of making the plots too busy.

vladsavelyev:
Putting low quality samples underneath actually sounds like a good idea!

vladsavelyev (approved):

PR #44, opened 2021-06-02
Add script to copy metadata to test reporting

vladsavelyev:


lgruen:
I believe this is obsolete with the new `metadata` buckets.

vladsavelyev:
Wanted to update it to copy the joint-calling meta.tsv file.

scripts/data_transfer/copy_test_data_for_reports.py
```diff
@@ -0,0 +1,37 @@
...
+is symmetrical to gs://cpg-tob-wgs-main/joint-calling/v2/meta.tsv and limited to samples in the
+test subset.
+"""
+
+import os
+import subprocess
+import click
+
+output = os.getenv('OUTPUT')
+assert output and output.startswith('gs://cpg-tob-wgs-test/')
```
lgruen:
```suggestion
assert output and output.startswith('gs://cpg-tob-wgs-main-metadata/')
```

scripts/data_transfer/copy_test_data_for_reports.py
```diff
@@ -0,0 +1,37 @@
...
+@click.option('--version', help='Joint-calling run version', required=True)
+def main(version):
+    """Entry point"""
+
+    subprocess.run(
+        [
+            'gsutil',
+            'cp',
+            f'gs://cpg-tob-wgs-main/joint-calling/{version}/meta.tsv',
+            f'gs://cpg-tob-wgs-main-metadata/joint-calling/{version}/meta.tsv',
```
lgruen: Did you mean to use `output` here?

scripts/data_transfer/copy_test_data_for_reports.py
```diff
@@ -0,0 +1,37 @@
...
+output = os.getenv('OUTPUT')
+assert output and output.startswith('gs://cpg-tob-wgs-test/')
+
+
+@click.command()
+@click.option('--version', help='Joint-calling run version', required=True)
+def main(version):
+    """Entry point"""
+
+    subprocess.run(
```
lgruen: Note that you can now also run simple bash scripts in the analysis-runner.

scripts/data_transfer/copy_test_data_for_reports.py
```diff
@@ -0,0 +1,37 @@
+"""
+Copy a sample-QC metadata TSV file generated by the joint-calling pipeline into the test bucket.
+The idea is that for interactive exploration of QC summary metrics and design of plots, it's safe
+to allow using the full set of samples, because the metadata doesn't contain any actual genomics
+variantion data. 
+
+Note that the test bucket also contains gs://cpg-tob-wgs-test/joint-calling/v2/meta.tsv, which
```
lgruen: Should probably be updated to refer to `test-metadata` now.

lgruen (changes requested):

scripts/data_transfer/copy_test_data_for_reports.py
vladsavelyev: Good point, a bash script would better suit here

scripts/data_transfer/copy_test_data_for_reports.py
vladsavelyev: But since we don't want the destination to be configurable, why use a variable here?

vladsavelyev:
Replaced with a bash script :)

scripts/data_transfer/copy_test_data_for_reports.py
lgruen: Well, we also copy a metadata file to `output` in the analysis-runner that describes that particular run. If you're not actually writing to that location, these destinations can be out of sync.

scripts/data_transfer/copy_joint_calling_metadata.sh
```diff
@@ -0,0 +1,12 @@
...
+VERSION=$1
+
+if [[ -z "$VERSION" ]]; then
+    echo "Provide a version tag of a joint-calling run as a first argument" 1>&2
+    exit 1
+fi
+
+gsutil cp \
+	gs://cpg-tob-wgs-main/joint-calling/${VERSION}/meta.tsv \
+	gs://cpg-tob-wgs-main-metadata/joint-calling/${VERSION}/meta.tsv
```
lgruen: Please see my other comment about potentially still using `$OUTPUT` here.

lgruen (approved):

PR #52, opened 2021-06-16
run concordance on v2-raw.mt

pdiakumis:
Took 14 minutes on 190 samples for v1-raw.mt (just chr22) - report at https://main-web.populationgenomics.org.au/tob-wgs/concordance/v1/v1-raw_chr22.html. Now will include those extra 94 (I think) samples in `v2-raw.mt`.


KatalinaBobowik (approved):
Looks good

PR #61, opened 2021-07-01
Copy 20 GVCFs to test. Remove old transfer scripts

vladsavelyev:


scripts/data_transfer/copy_test_gvcfs.py
```diff
@@ -5,25 +5,47 @@
...
+    'TOB1534',
+    'TOB1535',
+    'TOB1536',
+    'TOB1537',
+    'TOB1538',
+    'TOB1640',
+]
+
+for sn in SAMPLES:
+    subprocess.run(
```
lgruen, line 34: Note that you can now also run shell scripts directly, which might be a bit more convenient for tasks like these.

Doesn't matter much here, but if you use `gsutil -m cp`, you can copy in parallel, whereas this `for` loop will do things sequentially.

lgruen (approved):

scripts/data_transfer/copy_test_gvcfs.py
vladsavelyev: Nice, I knew about `gsutil -m cp` but haven't really used it much. Should be definitely useful for other scenarios.

PR #63, opened 2021-07-01
Update joint-calling fork, fix `--output-dir` analysis runner param

vladsavelyev:


pdiakumis (approved):
I had no idea there was a `gs://cpg-tob-wgs-hail` bucket too

PR #66, opened 2021-07-07
SNPchip ID exploration

pdiakumis:
See https://test-web.populationgenomics.org.au/tob-wgs/pdiakumis/snpchip_id_report/snpchip_ids.html for a test output.

@KatalinaBobowik this is also an example of using the analysis-runner to run an RMarkdown notebook using an R script to render it. You'll notice I'm installing additional packages into the conda environment (this job runs directly on the driver VM). 

I've also added R equivalents to the `bucket_path` and `output_path` functions you're familiar with from the Python runner scripts.


KatalinaBobowik (approved):
Looks great!

PR #69, opened 2021-07-20
Move uploaded SNPchip Plink files

pdiakumis:
@lgruen not sure if this is okay - copying to `test` then permanently moving to `main` from `upload`.


scripts/data_transfer/move_uploaded_snpchip_plink.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -ex
+
+INPUT_PATH="gs://cpg-tob-wgs-upload/one1k1_genotyping_all_regions"
+OUTPUT_PATH="snpchip/v1/plink"
+DESCRIPTION="OneK1K SNPchip data in Plink format, provided 2021-Jul-20 by GWCCG."
+
+# Copy files to 'test' for exploration with Hail.
+gsutil -m cp -r ${INPUT_PATH} gs://cpg-tob-wgs-test/${OUTPUT_PATH}
```
lgruen: Can we subset the data to correspond to the subset of samples that we also store for the joint calling tables etc. in `test`?

Ideally we'd mirror the _structure_ between `test` and `main`, but not the _content_.

(Didn't you also subset the previous SNP-chip data in a similar way?)

lgruen (changes requested):

scripts/data_transfer/move_uploaded_snpchip_plink.sh
pdiakumis: I'm just copying this over to main, skipping the Hail exploration in `test`. Or we can just keep in `main-upload`.

lgruen: Are you saying we're not planning on using this data at all for further analyses? If so, we could consider moving it to `archive`. If not, then we should have a subset in `test`. `upload` is only a staging area, not meant for long-term use.

pdiakumis: I'm not sure if we'll need these for further analyses (we probably will). The `bed` one below is a binary file output by Plink that contains the genotype calls. It needs the other two (`bim` and `fam`). There's that issue with the switched REF/ALT alleles, that's why I have not generated a MatrixTable which we could subset instead.

```
$ gsutil ls gs://cpg-tob-wgs-main-upload/one1k1_genotyping_all_regions/
gs://cpg-tob-wgs-main-upload/one1k1_genotyping_all_regions/onek1k.bed
gs://cpg-tob-wgs-main-upload/one1k1_genotyping_all_regions/onek1k.bim
gs://cpg-tob-wgs-main-upload/one1k1_genotyping_all_regions/onek1k.fam
```

lgruen: Is it possible to subset these with Plink directly?

pdiakumis: Yep, I can select 19 samples and also subset to chromosome e.g. 22?

lgruen: Awesome! Maybe we can keep all chromosomes, but use exactly the same samples that we also use as a subset for WGS? That would make it easy to combine the two datasets in tests.

scripts/data_transfer/move_uploaded_snpchip_plink.sh
```diff
@@ -0,0 +1,59 @@
...
+# and transfer a subset of 20 samples to test.
+
+set -ex
+
+# install Plink
+micromamba install -y --prefix $MAMBA_ROOT_PREFIX -y -c bioconda -c conda-forge plink
+
+SAMPLES_FILE="samples_to_keep.txt"
+
+cat > ${SAMPLES_FILE} << EOF
```
lgruen, line 16: What do these numbers mean? Is this the sample sample subset that we use for WGS joint calling?

pdiakumis: Those are the SNPchip Family + Individual IDs for those 20 samples in mt4. That's how Plink subsets by sample.

lgruen: Could you add a comment in the code describing that, please? (In general that's preferable to a response in a review comment, as the next person looking at this code might have a similar question (and won't review PR comments).)

pdiakumis: Ah, okay, will do!

lgruen (approved):
Thank you!

PR #72, opened 2021-07-27
Check sex labels

pdiakumis:
Checking sex labels between newly provided SNPchip Plink fam file and the original sex metadata file. See <https://test-web.populationgenomics.org.au/tob-wgs/pdiakumis/snpchip_plink_sex/check_sex1.html>. In the Results table, `Sex_cpg` is from the original sex metadata file, and `Sex` is from the Plink fam file (blank cells are missing data).


vladsavelyev (approved):
Awesome!

PR #78, opened 2021-11-15
QC report updates

vladsavelyev:
* Add `new_batches` parameter to specify which batches we need to highlight separately, highlight them in metrics plots and plot separately in per-batch plots. E.g.:
```
Rscript -e "rmarkdown::render('qc-for-html.Rmd', output_file='qc.html', params=list(meta_tsv='work/meta.tsv', namespace='main', new_batches='PBMC,Pilot/bone marrow'))"
```
* Sex plot: add `max.overlaps = NA` to `geom_text_repel` to label all bad samples. Highlight sex aneuploidies and ambiguous sex.
* Print nice tables for sex QC.
* Use `reactable` for all tables instead of printing raw data frames.

Example: https://main-web.populationgenomics.org.au/tob-wgs/qc/qc-v7.html


KatalinaBobowik (approved):
👍

PR #80, opened 2022-02-23
Add file to generate peer factor

KatalinaBobowik:
This PR is moved from the the [ancestry repository](https://github.com/populationgenomics/ancestry) (see the history [here](https://github.com/populationgenomics/ancestry/commits/peer/scripts/hail_batch/peer)).

### DIFF
--- a/scripts/peer/peer.py
+++ b/scripts/peer/peer.py
@@ -159,13 +159,13 @@ EOT"""
     )
 
     job.command(
-        f'python run_peer.py \
+        f"""python run_peer.py \
             {expression_file} \
             {covariates_file} \
             {job.factors_output_path} \
             {job.weights_output_path} \
             {job.precision_output_path} \
-            {job.residuals_output_path}'
+            {job.residuals_output_path}"""
     )
 
     return job
--- a/scripts/peer/peer.py
+++ b/scripts/peer/peer.py
@@ -53,7 +53,7 @@ def get_covariates(
     # 937
     # 69 samples are missing RNA-seq data
     # remove samples which don't have a OneK1K ID and only keep PCA scores
-    scores = scores[scores.OneK1K_ID.isna() == False][
+    scores = scores[~scores.OneK1K_ID.isna()][
         ['PC1', 'PC2', 'PC3', 'PC4', 'OneK1K_ID']
     ]
     # Merge PCA scores with Seyhan's covariate data and drop Seyhan's RNA-seq pca scores
@@ -66,7 +66,7 @@ def get_covariates(
     print(covariates[covariates.sampleid.isna()])
     # OneK1K ID 26_26 does not have any covariate data. Remove this sample and drop
     # the OneK1K_ID, since it's redundant with sampleid
-    covariates = covariates[covariates.sampleid.isna() == False].drop(
+    covariates = covariates[~covariates.sampleid.isna()].drop(
         ['OneK1K_ID'], axis=1
     )
     # Match expression data to covariate data, since PEER needs expr and covs
@@ -96,7 +96,7 @@ def get_covariates(
     covariates[['sex', 'age']] = covariates[['sex', 'age']].astype(int)
 
     # return expression data and covariates
-    return covariates.to_csv(index=False), expression.to_csv(index=False)
+    return covariates.to_csv(index=False), expression.to_csv(index=False) # type: ignore
 
 
 def get_at_index(obj, idx):
--- a/scripts/peer/peer.py
+++ b/scripts/peer/peer.py
@@ -200,7 +200,6 @@ def process_cell_type_on_batch(
     covariates_csv = load_data.call(get_at_index, intermediate_tuple, 0).as_str()
     expression_csv = load_data.call(get_at_index, intermediate_tuple, 1).as_str()
     peer_job = batch.new_job(job_prefix + 'peer')
-    peer_job.cpu(4)
     run_peer_job(peer_job, expression_csv, covariates_csv)
 
     batch.write_output(
--- a/scripts/peer/peer.py
+++ b/scripts/peer/peer.py
@@ -1,7 +1,7 @@
 """
 Test run this with:
 
-    cd <dir where peer.py file is>
+    cd scripts/peer
     analysis-runner \
         --access-level test \
         --description 'Test run peer analysis' \
--- a/scripts/peer/peer.py
+++ b/scripts/peer/peer.py
@@ -96,7 +96,8 @@ def get_covariates(
 
 def get_at_index(obj, idx):
     """
-    get object index
+    Get the object at index, required to break up the result
+    from the get_covariates data within Hail Batch    
     """
 
     return obj[idx]
--- a/scripts/peer/peer.py
+++ b/scripts/peer/peer.py
@@ -173,9 +173,9 @@ def process_cell_type_on_batch(
     batch: hb.Batch,
     cell_type_name: str,
     expression_file,
-    scores_path=SCORES_PATH,
-    covariates_path=COVARIATES_PATH,
-    sample_id_keys_path=SAMPLE_ID_KEYS_PATH,
+    scores_path,
+    covariates_path,
+    sample_id_keys_path,
 ):
     """
     Run PEER calculation in hail batch
--- a/scripts/peer/peer.py
+++ b/scripts/peer/peer.py
@@ -42,16 +42,11 @@ def get_covariates(
     scores.insert(loc=0, column='sampleid', value=sampleid)
     # filter to only CPG samples
     scores = scores[scores.sampleid.str.contains('CPG')]
-    print(len(scores.index))
-    # 1006
     # change CPG IDs to One1K1K IDs
     sampleid_keys = pd.read_csv(sample_id_keys_path, sep=',')
     scores = pd.merge(
         scores, sampleid_keys, how='left', left_on='sampleid', right_on='CPG_ID'
     )
-    print(len(scores.OneK1K_ID.dropna()))
-    # 937
-    # 69 samples are missing RNA-seq data
     # remove samples which don't have a OneK1K ID and only keep PCA scores
     scores = scores[~scores.OneK1K_ID.isna()][
         ['PC1', 'PC2', 'PC3', 'PC4', 'OneK1K_ID']
@@ -63,8 +58,9 @@ def get_covariates(
         scores, covariates, how='left', left_on='OneK1K_ID', right_on='sampleid'
     ).drop(['pc1', 'pc2', 'pc3', 'pc4'], axis=1)
     # check whether there's any missing data in the covariate data
-    print(covariates[covariates.sampleid.isna()])
-    # OneK1K ID 26_26 does not have any covariate data. Remove this sample and drop
+    missing_cov_data = list(covariates[covariates.sampleid.isna()].OneK1K_ID)
+    logging.info(f'{missing_cov_data} is missing covariate data')
+    # Remove samples with no covariate data and drop
     # the OneK1K_ID, since it's redundant with sampleid
     covariates = covariates[~covariates.sampleid.isna()].drop(
         ['OneK1K_ID'], axis=1
@@ -75,9 +71,8 @@ def get_covariates(
     merged_expr_covs = pd.merge(
         expression, covariates, how='right', left_on='sampleid', right_on='sampleid'
     )
-    print(merged_expr_covs.shape[0] - merged_expr_covs.dropna().shape[0])
-    # 15 samples with no expression data, but do have covariate and genotype data
-    # This differs per celltype
+    samples_missing_expression = merged_expr_covs.shape[0] - merged_expr_covs.dropna().shape[0]
+    logging.info(f'{samples_missing_expression} samples are missing expression data')
     # remove any rows with NA values
     merged_expr_covs = merged_expr_covs.dropna()
     expression = merged_expr_covs.drop(
--- a/scripts/peer/peer.py
+++ b/scripts/peer/peer.py
@@ -173,8 +173,6 @@ EOT"""
     return job
 
 
-# @click.command()
-# @click.option('--expression-file')
 def process_cell_type_on_batch(
     batch: hb.Batch,
     cell_type_name: str,
--- a/scripts/peer/peer.py
+++ b/scripts/peer/peer.py
@@ -21,14 +21,10 @@ from cpg_utils.hail import remote_tmpdir, output_path
 from google.cloud import storage
 
 PEER_DOCKER = 'australia-southeast1-docker.pkg.dev/cpg-common/images/peer:1.3.2'
-SCORES_PATH = 'gs://cpg-tob-wgs-test/kat/pca/nfe_feb22/v0/scores.json'
-COVARIATES_PATH = 'gs://cpg-tob-wgs-test/scrna-seq/grch38_association_files/covariates_files/covariates.tsv'
-SAMPLE_ID_KEYS_PATH = 'gs://cpg-tob-wgs-test/scrna-seq/grch38_association_files/metadata/keys_metadata_sheet.csv'
-
 
 def get_covariates(
     scores_path, covariates_path, expression_file, sample_id_keys_path
-) -> str:
+) -> Tuple[str, str]:
     """
     Get covariate data by merging PCA scores with age and sex info.
     Only needs to be run once. This returns a TSV (as a string)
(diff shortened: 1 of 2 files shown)


scripts/peer/peer.py
```diff
@@ -0,0 +1,139 @@
...
+    ).drop(['pc1', 'pc2', 'pc3', 'pc4'], axis=1)
+    # merge together covariates df and scores
+    covariates = pd.merge(
+        covariates, scores, how='left', left_on='InternalID', right_on='sampleid'
+    ).drop(['OneK1K_ID', 'InternalID', 'ExternalID', 'sampleid_y'], axis=1)
+
+    return covariates.to_numpy()
+
+
+def run_peer(expression_file, covariates, factors_output_path):
```
KatalinaBobowik: This is currently failing because dill is not installed. See the failed log [here](https://batch.hail.populationgenomics.org.au/batches/8273/jobs/2).

illusional: Hoping https://github.com/populationgenomics/images/pull/4/files will unblock this PR.

illusional: Hoping https://github.com/populationgenomics/images/pull/4 will unblock this PR.

illusional:
Just noting this is blocked because Hail Batch is trying to run the python job in Python 3, but peer is not installed for python 3:

- Can we make hail batch run in python 2 for this job?
- Can we install Peer for python 3?

KatalinaBobowik:
@illusional I don't think we can install Peer for python 3. I tried to look into running hail batch in pyhon 2 by installing a python 2.7 conda environment, then trying to run hail batch. Unfortunately, this doesn't seem to work. Alternatively, I could run the script in R. Would this solve any problems?

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
+"""
+Test run this with:
+
+    cd <dir where peer.py file is>
```
illusional: Can probably update this now that the path has been decided:

```suggestion
    cd scripts/peer
```

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
...
+import click
+import hailtop.batch as hb
+import pandas as pd
+from cpg_utils.hail import remote_tmpdir, output_path
+from google.cloud import storage
+
+PEER_DOCKER = 'australia-southeast1-docker.pkg.dev/cpg-common/images/peer:1.3.2'
+SCORES_PATH = 'gs://cpg-tob-wgs-test/kat/pca/nfe_feb22/v0/scores.json'
+COVARIATES_PATH = 'gs://cpg-tob-wgs-test/scrna-seq/grch38_association_files/covariates_files/covariates.tsv'
+SAMPLE_ID_KEYS_PATH = 'gs://cpg-tob-wgs-test/scrna-seq/grch38_association_files/metadata/keys_metadata_sheet.csv'
```
illusional: These constants are good for testing, but it would probably be better to remove them, and always specify the files when you run this script.

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
...
+    # load in scores which have outliers removed
+    scores_df = pd.read_json(scores_path)
+    sampleid = scores_df.s
+    # only get the first 4 PCs, as indicated by scree plot
+    scores = pd.DataFrame(scores_df['scores'].to_list()).iloc[:, 0:4]
+    scores.columns = ['PC1', 'PC2', 'PC3', 'PC4']
+    scores.insert(loc=0, column='sampleid', value=sampleid)
+    # filter to only CPG samples
+    scores = scores[scores.sampleid.str.contains('CPG')]
+    print(len(scores.index))
```
illusional: Left-over debug print, otherwise these should get turned into proper logging with a message (and using the logging library).

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
...
+    # 1006
+    # change CPG IDs to One1K1K IDs
+    sampleid_keys = pd.read_csv(sample_id_keys_path, sep=',')
+    scores = pd.merge(
+        scores, sampleid_keys, how='left', left_on='sampleid', right_on='CPG_ID'
+    )
+    print(len(scores.OneK1K_ID.dropna()))
```
illusional: ditto, I'll stop highlighting them but there are a couple more.

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
...
+
+PEER_DOCKER = 'australia-southeast1-docker.pkg.dev/cpg-common/images/peer:1.3.2'
+SCORES_PATH = 'gs://cpg-tob-wgs-test/kat/pca/nfe_feb22/v0/scores.json'
+COVARIATES_PATH = 'gs://cpg-tob-wgs-test/scrna-seq/grch38_association_files/covariates_files/covariates.tsv'
+SAMPLE_ID_KEYS_PATH = 'gs://cpg-tob-wgs-test/scrna-seq/grch38_association_files/metadata/keys_metadata_sheet.csv'
+
+
+def get_covariates(
+    scores_path, covariates_path, expression_file, sample_id_keys_path
+) -> str:
```
illusional: This doesn't match the return type

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
...
+    expression.sampleid.equals(covariates.sampleid)
+    # True
+    # remove sampleid from both covariates and expression dfs
+    expression.drop('sampleid', axis=1, inplace=True)
+    covariates.drop('sampleid', axis=1, inplace=True)
+    # finally, make sure age and sex in covariate df are int
+    covariates[['sex', 'age']] = covariates[['sex', 'age']].astype(int)
+
+    # return expression data and covariates
+    return covariates.to_csv(index=False), expression.to_csv(index=False)  # type: ignore
```
illusional: Can we remove the type ignore statement - the initial type annotation in the function is incorrect (it should be Tuple[str, str]). Otherwise can we add a comment on why it's ignored?

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+def get_at_index(obj, idx):
+    """
+    get object index
```
illusional:
```suggestion
    Get the object at index, required to break up the result
    from the get_covariates data within Hail Batch    
```

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
...
+    """
+    Run peer analysis, except because peer is in Python2, we can't take
+    advantage of the python_job concept in Batch. Instead, we'll rely on
+    the two inputs `expression_file` and `covariates_file` to be TSV files.
+    """
+
+    job.image(PEER_DOCKER)
+
+    job.command(f'echo "expressions file:" && head {expression_file}')
+    job.command(f'echo "covariates file:" && head {covariates_file}')
```
illusional: Can remove these extra print debug statements now

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
...
+import peer
+import numpy as np
+import pandas as pd
+
+def run_peer(expression_file, covariates_file, factors_output_path, weights_output_path, precision_output_path, residuals_output_path):
+    \"""
+    Get covariate data for each cell type
+    \"""
+
+    print 'Loading data'
```
illusional: Should remove these debug statements, or make them more descriptive

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
...
+    precision = model.getAlpha()
+    np.savetxt(precision_output_path, precision, delimiter=',')
+    # Calculate and save the residuals
+    residuals = model.getResiduals()
+    np.savetxt(residuals_output_path, residuals, delimiter=',')
+
+
+if __name__ == "__main__":
+    import sys
+    run_peer(sys.argv[1], sys.argv[2], sys.argv[3], sys.argv[4], sys.argv[5], sys.argv[6])
```
illusional: Maybe could simplify this with:

```suggestion
    run_peer(*sys.argv[1:])
```

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
...
+
+EOT"""
+    )
+
+    job.command(
+        f'python run_peer.py {expression_file} {covariates_file} {job.factors_output_path} {job.weights_output_path} {job.precision_output_path} {job.residuals_output_path}'
+    )
+    job.command('ls -l')
```
illusional: Debug statement to remove

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
...
+
+if __name__ == "__main__":
+    import sys
+    run_peer(sys.argv[1], sys.argv[2], sys.argv[3], sys.argv[4], sys.argv[5], sys.argv[6])
+
+EOT"""
+    )
+
+    job.command(
+        f'python run_peer.py {expression_file} {covariates_file} {job.factors_output_path} {job.weights_output_path} {job.precision_output_path} {job.residuals_output_path}'
```
illusional: It's probably worth breaking these up on different lines, you won't run into line length issues, and it'll be easier to debug the command in Hail Batch:

```suggestion
        f'python run_peer.py \
            {expression_file} \
            {covariates_file} \
            {job.factors_output_path} \
            {job.weights_output_path} \
            {job.precision_output_path} \
            {job.residuals_output_path}'
```

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
...
+
+# @click.command()
+# @click.option('--expression-file')
+def process_cell_type_on_batch(
+    batch: hb.Batch,
+    cell_type_name: str,
+    expression_file,
+    scores_path=SCORES_PATH,
+    covariates_path=COVARIATES_PATH,
+    sample_id_keys_path=SAMPLE_ID_KEYS_PATH,
```
illusional: As mentioned before, I think these should not be hard-coded anymore, but always supplied.

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
...
+
+def find_cell_types_from_path(path_to_cell_files):
+    """
+    Get cell types from expression data path
+    """
+
+    assert path_to_cell_files.startswith('gs://')
+    cs_client = storage.Client()
+
+    bucket_name = path_to_cell_files.split('gs://')[1].split('/')[0]
```
illusional, line 221:
```suggestion
    bucket_name = path_to_cell_files.lstrip('gs://').split('/')[0]
```

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
...
+    bucket = cs_client.get_bucket(bucket_name)
+    bucket_path = path_to_cell_files.split(f'gs://{bucket_name}/')[-1]
```
illusional, line 223: Another couple of ways that don't rely on split:

```suggestion
    bucket_path = path_to_cell_files.lstrip('gs://{bucket_name}/')
    bucket_path = path_to_cell_files[(5 + len(bucket_name) + 1):]
```

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
...
+
+    logging.info(f'Going to fetch cell types from {bucket_path}')
+    blobs = bucket.list_blobs(prefix=bucket_path)
```
illusional: This is important as otherwise it gets all files in all subdirectories of the path.

```suggestion
    blobs = bucket.list_blobs(prefix=bucket_path, delimiter='/')
```

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
...
+    job.command(
+        f'python run_peer.py {expression_file} {covariates_file} {job.factors_output_path} {job.weights_output_path} {job.precision_output_path} {job.residuals_output_path}'
+    )
+    job.command('ls -l')
+
+    return job
+
+
+# @click.command()
+# @click.option('--expression-file')
```
illusional:
```suggestion
```

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
...
+    """
+    This function finds all the cell types, and runs the inner
+    workflow for each cell_type, by constructing the path to the
+    expression files and the path to the covariates files.
+
+
+    """
+
+    dataset = os.getenv('CPG_DATASET')
+    driver_image = os.getenv('CPG_DRIVER_IMAGE')
```
illusional, line 253: You'll want an assert here to make sure the driver_image is set:
```suggestion
    driver_image = os.getenv('CPG_DRIVER_IMAGE')
    assert driver_image
```

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
...
+    backend = hb.ServiceBackend(billing_project=dataset, remote_tmpdir=remote_tmpdir())
+    batch = hb.Batch(name='PEER', backend=backend, default_python_image=driver_image)
+    cell_types: list = find_cell_types_from_path(path_to_cell_files)
+
+    for cell_type in cell_types:
+        expression_file = f'gs://cpg-tob-wgs-test/scrna-seq/grch38_association_files/expression_files/{cell_type}_expression.tsv'
```
illusional: Can you use the `path_to_cell_files` to construct this path - otherwise this should be passed in as an input to the script (this is linking to the test bucket atm)

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
...
+        process_cell_type_on_batch(batch, cell_type, expression_file=expression_file)
+
+        # batch.run(dry_run=True)
```
illusional:
```suggestion
```

scripts/peer/peer.py
```diff
@@ -0,0 +1,273 @@
...
+        if b.name.endswith('_expression.tsv')
+    ]
+    logging.info(f'Found {len(cell_types)} cell types: {cell_types}')
+
+    return cell_types
+
+
+@click.command()
+@click.option('--path-to-cell-files')
+def main(path_to_cell_files):
```
illusional: You'll now want to add an option for each scores_path, covariates_path, sample_id_keys_path, and connect them when you call `process_cell_type_on_batch` below.

illusional (changes requested):
There are a few debug statements left in the script, and a couple of variables that IMO should be parameterised, but looks really great!

scripts/peer/peer.py
KatalinaBobowik: deleted the print message

scripts/peer/peer.py
KatalinaBobowik: I've used the logging library as well

scripts/peer/peer.py
KatalinaBobowik: Maybe I missed it, but why should we be avoiding using `split`?

scripts/peer/peer.py
KatalinaBobowik: I'm a little confused on what would go here instead, since we have two dataframes being returned

KatalinaBobowik: as you mentioned below, would it look like this?
`-> Tuple[str, str]`

scripts/peer/peer.py
KatalinaBobowik: I've removed the type ignore statement after updating the return type

scripts/peer/peer.py
KatalinaBobowik: I removed them

scripts/peer/peer.py
KatalinaBobowik: removed

scripts/peer/peer.py
KatalinaBobowik: good catch, added `path_to_cell_files` instead

scripts/peer/peer.py
KatalinaBobowik: I used the second example but am getting the error:
`Using .strip() with multi-character strings is misleading the reader. It looks like stripping a substring. Move your character set to a constant if this is deliberate. Use .replace() or regular expressions to remove string fragments.`

KatalinaBobowik: (from pylint)

scripts/peer/peer.py
KatalinaBobowik: I'm getting a pylint error:
`undefined name 'Tuple'`. Not sure if this should be lowercase, or something else?

illusional: You'll need to import it from typing:

```python
from typing import Tuple
```

scripts/peer/peer.py
illusional: Ah okay, let's not do that :) `.strip` actually doesn't work in the way I expected it to. Could you revert this back to your original code?

KatalinaBobowik: np, will do that now

scripts/peer/peer.py
```diff
@@ -0,0 +1,272 @@
...
+    )
+
+    job.command(
+        f'python run_peer.py \
+            {expression_file} \
+            {covariates_file} \
+            {job.factors_output_path} \
+            {job.weights_output_path} \
+            {job.precision_output_path} \
+            {job.residuals_output_path}'
```
illusional: My suggestion here wasn't correct, this should be a multiline string (rather than just adding new lines), sorry this suggestion should be better:

```suggestion
        f"""python run_peer.py \
            {expression_file} \
            {covariates_file} \
            {job.factors_output_path} \
            {job.weights_output_path} \
            {job.precision_output_path} \
            {job.residuals_output_path}"""
```

illusional (approved):

PR #81, opened 2022-03-07
Generate genotype df

KatalinaBobowik:
This generates a genotype file to Plink format to be used in the association analysis. Only  variants with an MAF > 0.01 are retained + additional filtering for passing VQSR filters, GQ, and missingness. This has run [successfully](https://batch.hail.populationgenomics.org.au/batches/8390/jobs/2) in the `test` bucket.

### DIFF
--- a/scripts/generate_genotype_df/README.md
+++ b/scripts/generate_genotype_df/README.md
@@ -1,6 +1,6 @@
 # Export TOB-WGS dataset to PLINK format
 
-This runs a Hail query script in Dataproc using Hail Batch in order to export the TOB-WGS genotypes, which are requirements for the association analysis. Only variants that pass VQSR and GQ filters are retained, as well as variants with an MAF > 0.01. To run, use conda to install the analysis-runner, then execute the following command:
+This runs a Hail query script in Dataproc using Hail Batch in order to export the TOB-WGS genotypes, which are requirements for the association analysis. Only variants that pass VQSR and GQ filters are retained, as well as variants with an MAF > 0.01. To run, use pip to install the analysis-runner, then execute the following command:
 
 ```sh
 analysis-runner --dataset tob-wgs \
--- a/scripts/generate_genotype_df/README.md
+++ b/scripts/generate_genotype_df/README.md
@@ -1,0 +1,9 @@
+# Export TOB-WGS dataset to PLINK format
+
+This runs a Hail query script in Dataproc using Hail Batch in order to export the TOB-WGS genotypes, which are requirements for the association analysis. Only variants that pass VQSR and GQ filters are retained, as well as variants with an MAF > 0.01. To run, use conda to install the analysis-runner, then execute the following command:
+
+```sh
+analysis-runner --dataset tob-wgs \
+--access-level standard --output-dir "tob_wgs_plink_maf01/v0" \
+--description "TOB plink" python3 main.py
+```
--- a/scripts/generate_genotype_df/generate_genotype_df.py
+++ b/scripts/generate_genotype_df/generate_genotype_df.py
@@ -25,7 +25,7 @@ def query():
     # filter out variants with MAF < 0.01
     mt = mt.filter_rows(mt.freq.AF[1] > 0.01)
     tob_wgs_path = output_path('tob_wgs_maf01.parquet')
-    mt.rows().to_spark().write.mode("append").parquet(tob_wgs_path)
+    mt.rows().to_spark().write.mode('append').parquet(tob_wgs_path)
 
 if __name__ == '__main__':
     query()
--- a/scripts/generate_genotype_df/generate_genotype_df.py
+++ b/scripts/generate_genotype_df/generate_genotype_df.py
@@ -3,23 +3,29 @@
 """Generate genotype dfs for the association analysis"""
 
 import hail as hl
+from analysis_runner import bucket_path, output_path
 
-TOB_WGS = 'gs://cpg-tob-wgs-test/mt/v7.mt/'
+TOB_WGS = bucket_path('mt/v7.mt/')
 
-hl.init(default_reference='GRCh38')
+def query():
+    """Query script entry point."""
 
-mt = hl.read_matrix_table(TOB_WGS)
-# filter out variants that didn't pass the VQSR filter
-mt = mt.filter_rows(hl.is_missing(mt.filters))
-# VQSR does not filter out low quality genotypes. Filter these out
-mt = mt.filter_entries(mt.GQ <= 20, keep = False)
-# filter out samples with a genotype call rate > 0.8 
-n_samples = mt.count_cols()
-call_rate = 0.8
-mt = mt.filter_rows(hl.agg.sum(hl.is_missing(mt.GT)) > (n_samples * call_rate), keep=False)
-# filter out variants with MAF < 0.01
-mt = mt.filter_rows(mt.freq.AF[1] > 0.01)
-# turn files into output that we need
-# turn into plink, split up by chromosome
+    hl.init(default_reference='GRCh38')
 
+    mt = hl.read_matrix_table(TOB_WGS)
+    # filter out variants that didn't pass the VQSR filter
+    mt = mt.filter_rows(hl.is_missing(mt.filters))
+    # VQSR does not filter out low quality genotypes. Filter these out
+    mt = mt.filter_entries(mt.GQ <= 20, keep = False)
+    # filter out samples with a genotype call rate > 0.8 
+    n_samples = mt.count_cols()
+    call_rate = 0.8
+    mt = mt.filter_rows(hl.agg.sum(hl.is_missing(mt.GT)) > (n_samples * call_rate), keep=False)
+    # filter out variants with MAF < 0.01
+    mt = mt.filter_rows(mt.freq.AF[1] > 0.01)
+    # export as PLINK file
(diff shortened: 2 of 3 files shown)


scripts/generate_genotype_df/generate_genotype_df.py
```diff
@@ -0,0 +1,31 @@
...
+TOB_WGS = bucket_path('mt/v7.mt/')
+
+def query():
+    """Query script entry point."""
+
+    hl.init(default_reference='GRCh38')
+
+    mt = hl.read_matrix_table(TOB_WGS)
+    # filter out variants that didn't pass the VQSR filter
+    mt = mt.filter_rows(hl.is_missing(mt.filters))
```
MattWellie: I thought Hail used an empty set for 0 applied filters? I've just tried this syntax in a notebook and all rows with no applied filters were removed.

This alternate syntax allows filters to be missing or a 0 length set, and will pass in both situations

```
matrix = matrix.filter_rows(
    hl.len(
        hl.or_else(
            matrix.filters, 
            hl.empty_set(hl.tstr)
        )
    ) == 0
)
```

vladsavelyev: In the joint-calling pipeline we have empty sets for column-level filters, and NA for row-level filters. Good catch - should unify that. 

But agree that it's better to always handle both situations using the code you shared. Thanks Matt!

vladsavelyev (approved):
Just +1 to Matt's suggestion, approving otherwise

scripts/generate_genotype_df/generate_genotype_df.py
```diff
@@ -0,0 +1,31 @@
...
+    mt = mt.filter_entries(mt.GQ <= 20, keep = False)
+    # filter out samples with a genotype call rate > 0.8 (as in the gnomAD supplementary paper)
+    n_samples = mt.count_cols()
+    call_rate = 0.8
+    mt = mt.filter_rows(hl.agg.sum(hl.is_missing(mt.GT)) > (n_samples * call_rate), keep=False)
+    # filter out variants with MAF < 0.01
+    mt = mt.filter_rows(mt.freq.AF[1] > 0.01)
+    # export as PLINK file
+    tob_wgs_path = output_path('tob_wgs_plink_maf01')
+    hl.export_plink(mt, tob_wgs_path, ind_id=mt.s)
```
lgruen: As discussed over Zoom, would it be an option to export the Hail Table as a Pandas DataFrame directly, e.g. using Parquet?

scripts/generate_genotype_df/main.py
```diff
@@ -0,0 +1,23 @@
...
+    billing_project=os.getenv('HAIL_BILLING_PROJECT'), bucket=os.getenv('HAIL_BUCKET')
+)
+
+batch = hb.Batch(name='export_genotype_data', backend=service_backend)
+
+dataproc.hail_dataproc_job(
+    batch,
+    f'generate_genotype_df.py',
+    max_age='4h',
+    num_secondary_workers=20,
```
lgruen: Do you need that many workers and the disk override or is this just a remnant from a previous script?

lgruen (changes requested):

scripts/generate_genotype_df/generate_genotype_df.py
KatalinaBobowik: that's really helpful, thanks @MattWellie!

scripts/generate_genotype_df/main.py
KatalinaBobowik: This was an old remnant- thanks for pointing it out!

scripts/generate_genotype_df/generate_genotype_df.py
KatalinaBobowik: I've exported this to parquet now

scripts/generate_genotype_df/README.md
```diff
@@ -0,0 +1,9 @@
+# Export TOB-WGS dataset to PLINK format
+
+This runs a Hail query script in Dataproc using Hail Batch in order to export the TOB-WGS genotypes, which are requirements for the association analysis. Only variants that pass VQSR and GQ filters are retained, as well as variants with an MAF > 0.01. To run, use conda to install the analysis-runner, then execute the following command:
```
lgruen: Nit: the analysis-runner is now installed using pip.

scripts/generate_genotype_df/generate_genotype_df.py
```diff
@@ -0,0 +1,32 @@
...
+    """Query script entry point."""
+
+    hl.init(default_reference='GRCh38')
+
+    mt = hl.read_matrix_table(TOB_WGS)
+    # filter out variants that didn't pass the VQSR filter
+    mt = mt.filter_rows(hl.len(hl.or_else(mt.filters, hl.empty_set(hl.tstr))) == 0)
+
+    # VQSR does not filter out low quality genotypes. Filter these out
+    mt = mt.filter_entries(mt.GQ <= 20, keep = False)
```
lgruen:
```suggestion
    mt = mt.filter_entries(mt.GQ <= 20, keep=False)
```

scripts/generate_genotype_df/generate_genotype_df.py
```diff
@@ -0,0 +1,32 @@
...
+    # filter out samples with a genotype call rate > 0.8 (as in the gnomAD supplementary paper)
+    n_samples = mt.count_cols()
+    call_rate = 0.8
+    mt = mt.filter_rows(hl.agg.sum(hl.is_missing(mt.GT)) > (n_samples * call_rate), keep=False)
+    # filter out variants with MAF < 0.01
+    mt = mt.filter_rows(mt.freq.AF[1] > 0.01)
+    # export as PLINK file
```
lgruen:
```suggestion
```

scripts/generate_genotype_df/generate_genotype_df.py
```diff
@@ -0,0 +1,32 @@
...
+    tob_wgs_path = output_path('tob_wgs_maf01.parquet')
+    mt.rows().to_spark().write.mode("append").parquet(tob_wgs_path)
```
lgruen: This will write a Spark DataFrame, not a Pandas DataFrame. Are you sure that's what you need downstream?

lgruen:
```suggestion
    mt.rows().to_spark().write.mode("append").parquet(tob_wgs_path)
```
```suggestion
    mt.rows().to_spark().write.mode('append').parquet(tob_wgs_path)
```

scripts/generate_genotype_df/main.py
```diff
@@ -0,0 +1,21 @@
...
+)
+
+batch = hb.Batch(name='export_genotype_data', backend=service_backend)
+
+dataproc.hail_dataproc_job(
+    batch,
+    f'generate_genotype_df.py',
+    max_age='4h',
+    init=['gs://cpg-reference/hail_dataproc/install_common.sh'],
+    job_name=f'export_genotype_data',
```
lgruen:
```suggestion
    job_name='export_genotype_data',
```

scripts/generate_genotype_df/main.py
```diff
@@ -0,0 +1,21 @@
...
+
+service_backend = hb.ServiceBackend(
+    billing_project=os.getenv('HAIL_BILLING_PROJECT'), bucket=os.getenv('HAIL_BUCKET')
+)
+
+batch = hb.Batch(name='export_genotype_data', backend=service_backend)
+
+dataproc.hail_dataproc_job(
+    batch,
+    f'generate_genotype_df.py',
```
lgruen:
```suggestion
    'generate_genotype_df.py',
```

lgruen (changes requested):

scripts/generate_genotype_df/generate_genotype_df.py
KatalinaBobowik: This saves just fine as a parquet file that I can load in as a pandas df with `pd.read_parquet`. Is there a better way to do this?

scripts/generate_genotype_df/README.md
KatalinaBobowik: Fixed

scripts/generate_genotype_df/generate_genotype_df.py
lgruen: There's `to_pandas`, maybe that will have a slightly different format? @vladsaveliev, do you know?

lgruen (approved):
Added a note regarding `to_spark` vs `to_pandas`, which you might want to consider changing before merging.

scripts/generate_genotype_df/generate_genotype_df.py
KatalinaBobowik: ok, I tested this out and while both dfs are the same when reading back into python, the `to_spark` way results in a much larger file. I've now changed this to `to_pandas` (and only take the fields I need), and then export to parquet

lgruen: Cool, thanks! You probably also don't need the "append" write mode here, right?

KatalinaBobowik: No, I just did the `mt.rows().to_pandas().to_parquet()`

lgruen: Perfect! 😸

PR #83, opened 2022-03-10
Filter to columns used in downstream analysis, convert to pandas dataframe, and split chromosomes

KatalinaBobowik:
Filter to columns used in downstream analysis, convert to pandas dataframe, and save each chromosome as individual files. [Successfully](https://batch.hail.populationgenomics.org.au/batches/8682/jobs/2) run in `test`

### DIFF
--- a/scripts/generate_genotype_df/generate_genotype_df.py
+++ b/scripts/generate_genotype_df/generate_genotype_df.py
@@ -24,8 +24,9 @@ def query():
     mt = mt.filter_rows(hl.agg.sum(hl.is_missing(mt.GT)) > (n_samples * call_rate), keep=False)
     # filter out variants with MAF < 0.01
     mt = mt.filter_rows(mt.freq.AF[1] > 0.01)
-    tob_wgs_path = output_path('tob_wgs_maf01.parquet')
-    mt.rows().to_pandas()[['locus.contig','locus.position','alleles']].to_parquet(tob_wgs_path)
+    pd = mt.rows().to_pandas()[['locus.contig','locus.position','alleles']]
+    for chr in set(pd['locus.contig']): 
+        pd.loc[pd['locus.contig'] == chr].to_parquet(output_path(f'tob_genotype_maf01_{chr}.parquet'))
 
 if __name__ == '__main__':
     query()
--- a/scripts/generate_genotype_df/generate_genotype_df.py
+++ b/scripts/generate_genotype_df/generate_genotype_df.py
@@ -27,9 +27,10 @@ def query():
     mt = mt.filter_rows(mt.freq.AF[1] > 0.01)
     # select only locus and alleles, which are the keys, then convert to pandas
     pd = mt.rows().select().to_pandas()
+    print(pd.head())
     # save each chromosome to an individual file
-    for chr in set(pd['locus.contig']): 
-        pd.loc[pd['locus.contig'] == chr].to_parquet(output_path(f'tob_genotype_maf01_{chr}.parquet'))
+    # for chr in set(pd['locus.contig']): 
+    #     pd.loc[pd['locus.contig'] == chr].to_parquet(output_path(f'tob_genotype_maf01_{chr}.parquet'))
 
 if __name__ == '__main__':
     query()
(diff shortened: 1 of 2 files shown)


scripts/generate_genotype_df/generate_genotype_df.py
```diff
@@ -24,8 +25,15 @@ def query():
     mt = mt.filter_rows(hl.agg.sum(hl.is_missing(mt.GT)) > (n_samples * call_rate), keep=False)
     # filter out variants with MAF < 0.01
     mt = mt.filter_rows(mt.freq.AF[1] > 0.01)
-    tob_wgs_path = output_path('tob_wgs_maf01.parquet')
-    mt.rows().to_pandas()[['locus.contig','locus.position','alleles']].to_parquet(tob_wgs_path)
+    # select only locus and alleles, which are the keys, then convert to pandas
+    locus_alleles = mt.rows().select().to_pandas(flatten=True)
+    # expand locus to two columns and rename 
+    locus_alleles = locus_alleles.astype(str)
```
lgruen: Instead of converting `Locus` to a string and then pulling it apart using `split`, can you use the `contig` and `position` fields directly?
https://hail.is/docs/0.2/genetics/hail.genetics.Locus.html#hail.genetics.Locus

scripts/generate_genotype_df/generate_genotype_df.py
```diff
@@ -13,9 +14,9 @@ def query():
     hl.init(default_reference='GRCh38')
 
     mt = hl.read_matrix_table(TOB_WGS)
+    mt = hl.experimental.densify(mt)
```
lgruen, line 17: Do you need this? Won't this just affect the reference blocks?

lgruen (changes requested):

scripts/generate_genotype_df/generate_genotype_df.py
KatalinaBobowik: I thought densifying was necessary whenever you use the `GT` field?

scripts/generate_genotype_df/generate_genotype_df.py
MattWellie: I tried to do a matrix.locus.contig.collect_as_set() but the notebook ran out of memory!

KatalinaBobowik: @MattWellie I've never seen `collect_as_set()` before. I just tried it on a small subset now. 

@lgruen The issue I'll have if I use `mt.locus.contig` or `mt.locus.position` directly is that it gets converted to a string expression rather than staying as a hail table (which has the `to_pandas` functionality). The best I could come up with instead is something like this:
`mt = mt.annotate_rows(contig = mt.locus.contig, position=mt.locus.position).key_rows_by('alleles')
locus_alleles = mt.rows().select('contig','position').to_pandas(flatten=True)
`

I can't select rows by `mt.locus.contig` or `mt.locus.position` directly, unfortunately

scripts/generate_genotype_df/generate_genotype_df.py
lgruen: Sorry, of course you're right. I keep forgetting about the actual semantics of `densify`. Loic put it pretty succinctly a while ago: "densify does not add or remove rows, ever. It simply fills up missing entries." (https://nbviewer.org/github/populationgenomics/tob-deepdive/blob/main/densify.ipynb).

scripts/generate_genotype_df/generate_genotype_df.py
lgruen: Could you instead say:
```
t = mt.rows()
t = t.key_by(contig = t.locus.contig, position = t.locus.position)
t = t.select(t.alleles)
pd = t.to_pandas(flatten=True)
```
?

lgruen (changes requested):
Looks like there's also a merge conflict that you'll need to resolve first.

KatalinaBobowik:
@lgruen looking into the merge conflict now

scripts/generate_genotype_df/generate_genotype_df.py
KatalinaBobowik: that's a lot more readable. I've incorporated this version in. Thanks!

lgruen (approved):

PR #93, opened 2022-03-25
Export entries, rather than rows

KatalinaBobowik:
I need to have genotype data for each individual, rather than information for each allele. I've therefore converted the entries field to a hail table and manipulated this to be a sample x SNP df (with genotypes as values). Currently running in `test`

### DIFF
--- a/scripts/generate_genotype_df/generate_genotype_df.py
+++ b/scripts/generate_genotype_df/generate_genotype_df.py
@@ -36,15 +36,16 @@ def query():
     t = t.key_by(contig=t.locus.contig, position=t.locus.position)
     t = t.select(t.alleles, t.s, t.n_alt_alleles)
     df = t.to_pandas(flatten=True)
-    # create a new matrix with SNPs as columns and samples as rows
-    # with genotype values as entries
-    df['snp'] = df['contig'].map(str) + ':' + df['position'].map(str) + ':' + df['alleles'].str[0] + ':' + df['alleles'].str[1]
-    columns = (df.groupby(['s']).agg({'snp': lambda x: x.tolist()})).snp[0]
-    snps = (df.groupby(['s']).agg({'n_alt_alleles': lambda x: x.tolist()}))
-    df = pd.DataFrame(snps['n_alt_alleles'].to_list(), columns=columns, index=snps.index) 
     # save each chromosome to an individual file
     for chr in set(df['contig']):
-        df.loc[df['contig'] == chr].to_parquet(output_path(f'tob_genotype_maf01_{chr}.parquet'))
+        df = df.loc[df['contig'] == chr]
+        # create a new matrix with SNPs as columns and samples as rows
+        # with genotype values as entries
+        df['snp_id'] = df['contig'].map(str) + ':' + df['position'].map(str) + ':' + df['alleles'].str[0] + ':' + df['alleles'].str[1]
+        columns = (df.groupby(['s']).agg({'snp_id': lambda x: x.tolist()})).snp_id[0]
+        genotypes = (df.groupby(['s']).agg({'n_alt_alleles': lambda x: x.tolist()}))
+        df = pd.DataFrame(genotypes['n_alt_alleles'].to_list(), columns=columns, index=genotypes.index) 
+        df.to_parquet(output_path(f'tob_genotype_maf01_{chr}.parquet'))
 
 
 if __name__ == '__main__':
--- a/scripts/generate_genotype_df/generate_genotype_df.py
+++ b/scripts/generate_genotype_df/generate_genotype_df.py
@@ -3,6 +3,7 @@
 """Generate genotype dfs for the association analysis"""
 
 import hail as hl
+import pandas as pd
 from cpg_utils.hail import dataset_path, output_path
 
 TOB_WGS = dataset_path('mt/v7.mt/')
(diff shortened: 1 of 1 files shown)


scripts/generate_genotype_df/generate_genotype_df.py
```diff
@@ -29,15 +30,22 @@ def query():
...
+    t = t.select(t.alleles, t.s, t.n_alt_alleles)
+    df = t.to_pandas(flatten=True)
     # save each chromosome to an individual file
-    for chr in set(pd['contig']): 
-        pd.loc[pd['contig'] == chr].to_parquet(output_path(f'tob_genotype_maf01_{chr}.parquet'))
+    for chr in set(df['contig']):
+        df = df.loc[df['contig'] == chr]
+        # create a new matrix with SNPs as columns and samples as rows
+        # with genotype values as entries
+        df['snp_id'] = df['contig'].map(str) + ':' + df['position'].map(str) + ':' + df['alleles'].str[0] + ':' + df['alleles'].str[1]
```
lgruen, line 44: Didn't we think recently that SNP ID isn't needed? Why not export these individual fields as different columns?

scripts/generate_genotype_df/generate_genotype_df.py
```diff
@@ -29,15 +30,22 @@ def query():
...
+        columns = (df.groupby(['s']).agg({'snp_id': lambda x: x.tolist()})).snp_id[0]
```
lgruen, line 45: There's probably a more efficient way to do this (if you really need it), based on the original MT. Probably worth checking with @vladsaveliev and looking at the code that exports MTs to VCFs.

scripts/generate_genotype_df/generate_genotype_df.py
```diff
@@ -29,15 +30,22 @@ def query():
...
-    # then convert to pandas
-    t = mt.rows()
+    t = mt.entries()
+    # get genotype codes (n_alt_alleles, as in PLINK files)
+    t = t.annotate(n_alt_alleles=t.GT.n_alt_alleles())
     t = t.key_by(contig=t.locus.contig, position=t.locus.position)
-    t = t.select(t.alleles)
-    pd = t.to_pandas(flatten=True)
-    # expand locus to two columns and rename 
+    t = t.select(t.alleles, t.s, t.n_alt_alleles)
```
lgruen, line 37: This will blow up the data frame _significantly_. This also won't scale well with larger number of samples. Maybe it's time to reconsider whether we shouldn't be implementing this all in Hail instead, even if that means we need to find an alternative to the Spearman correlation computation?

lgruen (changes requested):

scripts/generate_genotype_df/generate_genotype_df.py
KatalinaBobowik: I'm trying to get a matrix which holds genotype information for each sample across all SNPs. The SNP ID wouldn't be needed if you were only looking at allele information, but across genotype information, it's the only way to have a row x column matrix. e.g., 

| Sample  | SNP1 | SNP2 | SNP3 |
|---------| ------|-------|------|
| sample1 | 1  | 0 | 0 | 1 |
| sample2 | 2  | 1 | 2 | 1 |
| sample3 | 0  | 0 | 2 | 1 |

scripts/generate_genotype_df/generate_genotype_df.py
KatalinaBobowik: Would it be easier to just export the PLINK files, as done originally?

lgruen: That won't really change the problem that you'd need to hold this complete table (or at least the section for one chromosome) in memory in a single node though, so fundamentally this seems to be very limited in terms of scalability.

I think if you really wanted to do this outside of Hail, you'd need to generate a DataFrame for much smaller regions and export them as files individually. Then you could run the association analysis on those individual chunks separately.

At that point you're kind of reimplementing some of the parallelization that Hail would take care of for you directly.

lgruen (changes requested):
Added a comment about potentially chunking the export in a different fashion, to make this more tractable. I'd still suggest considering whether we can't run this within Hail, potentially using a different type of correlation algorithm?

scripts/generate_genotype_df/generate_genotype_df.py
KatalinaBobowik: I'm struggling to understand how this is so different from the way we ran the script previously, where we exported a parquet file for each chromosome?

lgruen: Didn't you previously only export site-level data, i.e. not per-sample genotypes?

I'm not saying that splitting into chunks is fundamentally different from splitting into chromosomes, but the amount of memory required per job is just way smaller if you can have much more fine-grained chunks.

For example, instead of exporting a file per chromosome here, could you instead launch 1000 jobs that each open the Hail MT, but process a chunk that's 1/1000 of the whole genome? Or maybe have a chunk per gene (with 1MB each side) or similar?

lgruen: (You could still do the processing using a Pandas DataFrame that way, but you'd skip exporting in a single-threaded way in this script completely.)

KatalinaBobowik: Ok, I completely overlooked that this inflated the size so much. I'll look into chunking this

PR #95, opened 2022-05-11
Eqtl hail batch

KatalinaBobowik:


scripts/eqtl_hail_batch/launch_conditional_analysis.py
```diff
@@ -0,0 +1,160 @@
...
+
+    bucket_name = input_path.split('gs://')[1].split('/')[0]
+    bucket = cs_client.get_bucket(bucket_name)
+    bucket_path = input_path.split(f'gs://{bucket_name}/')[-1]
+
+    def file_exists(files):
+        file_status = storage.Blob(
+            bucket=bucket, name=files.split(bucket_name)[-1].strip('/')
+        ).exists(cs_client)
+        return file_status
```
MattWellie, line 86: This is a good use case for cloudpathlib!

you can do a cheeky 
`from cloudpathlib import AnyPath`

then 
`if AnyPath(file).exists(): ... `

MattWellie: https://cloudpathlib.drivendata.org/stable/#quick-usage
This can also be used to list contents of a local or remote directory!

```
p = list(AnyPath('gs://cpg-acute-care-test/reanalysis').glob('*'))
p[0].exists()
```

scripts/eqtl_hail_batch/launch_generate_eqtl_spearman.py
```diff
@@ -0,0 +1,149 @@
...
+        # we can infer the cell types from the 'expression_files'
+        # subdirectory of the input_path
+        # eg: {cell_type}_expression.tsv
+
+        path_to_expression_files = os.path.join(bucket_path, 'expression_files')
+        logging.info(f'Going to fetch cell types from {path_to_expression_files}')
+        blobs = bucket.list_blobs(prefix=path_to_expression_files + '/', delimiter='/')
+
+        cell_types = [
+            os.path.basename(b.name)[:-15]
```
vladsavelyev: [nitpicking, feel free to ignore]

15 is the length of `_expression.tsv`? Could be error-prone. I'd assign `_expression.tsv` to a variable and use `len()`, e.g.:

```
ending = '_expression.tsv'
os.path.basename(b.name)[:-len(ending)]
```

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -0,0 +1,448 @@
...
+    batch = hb.Batch(name='eQTL', backend=backend, default_python_image=DRIVER_IMAGE)
+
+    # load in files to get number of scatters and residuals
+    expression_df_literal = pd.read_csv(AnyPath(expression), sep='\t')
+    geneloc_df_literal = pd.read_csv(AnyPath(geneloc), sep='\t')
+    covariate_df_literal = pd.read_csv(AnyPath(covariates), sep=',')
+
+    # only run batch jobs if there are test genes in the chromosome
+    n_genes_in_scatter = get_number_of_scatters(expression_df_literal, geneloc_df_literal)
+    if n_genes_in_scatter > 0:
```
vladsavelyev: [nitpicking]
In what situations it can equal 0? Wonder it makes sense to print some error message to explicitly catch invalid inputs.

Also, in terms of code style, from my opinion it's preferable to avoid unnecessary indentation if possible, so instead of:

```
if n_genes_in_scatter > 0:
    <a lot of code>
```

Better do:

```
if n_genes_in_scatter <= 0:
    raise ValueError('')

<a lot of code>
```

scripts/eqtl_hail_batch/README.md
```diff
@@ -0,0 +1,45 @@
+# Run hail batch job to generate list of eQTLs
+
+This runs a Hail batch script in order to generate a list of eQTLs from scRNA-seq expression. This code was taken from Seyhan Yazar from Joseph Powell's group at the Garvan-Weizmann Centre for Cellular Genomics, then converted into Python/hail batch. To run, use conda to install the analysis-runner. For the first round of eQTLs, execute the following command for running one cell type and chromosome (B_intermediate cells and chromosome 22 are shown here as an example):
+
+```sh
+analysis-runner --dataset tob-wgs \
+    --access-level test --output-dir "plasma/chr22/v0"  --memory=highmem --cpu=8 \
```
illusional: Does the analysis-runner script need this much memory? Doesn't it just load stuff and move all the work to the hail batch workflow?

scripts/eqtl_hail_batch/launch_generate_eqtl_spearman.py
KatalinaBobowik: That's a really good point. Hard-coding is not ideal, so this is a much better way. I've updated this now. Thanks!

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
KatalinaBobowik: I've added in the `raise ValueError('')` message here and fixed the indentation. Good to know about the code style, I didn't know about that

scripts/eqtl_hail_batch/launch_conditional_analysis.py
KatalinaBobowik: Awesome, I didn't know you can do this in cloudpathlib. Does one have an advantage over another, or are these just different options?

scripts/eqtl_hail_batch/conditional_analysis.py
```diff
@@ -0,0 +1,467 @@
...
+    t = t.filter(set_to_keep.contains(t['snpid']))
+    # only keep SNPs where all samples have an alt_allele value
+    snps_to_remove = set(t.filter(hl.is_missing(t.n_alt_alleles)).snpid.collect())
+    if len(snps_to_remove) > 0:
+        t = t.filter(~hl.literal(snps_to_remove).contains(t.snpid))
+        genotype_df = t.to_pandas(flatten=True)
+        genotype_df.rename({'onek1k_id': 'sampleid'}, axis=1, inplace=True)
+    else:
+        genotype_df = t.to_pandas(flatten=True)
+        genotype_df.rename({'onek1k_id': 'sampleid'}, axis=1, inplace=True)
```
illusional:
```suggestion
    if len(snps_to_remove) > 0:
        t = t.filter(~hl.literal(snps_to_remove).contains(t.snpid))

    genotype_df = t.to_pandas(flatten=True)
    genotype_df.rename({'onek1k_id': 'sampleid'}, axis=1, inplace=True)
```

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -0,0 +1,449 @@
...
+
+def get_number_of_scatters(expression_df, geneloc_df):
+    """get index of total number of genes"""
+
+    expression_df = filter_lowly_expressed_genes(expression_df)
+    gene_ids = list(expression_df.columns.values)[1:]
+    geneloc_df = geneloc_df[geneloc_df.gene_name.isin(gene_ids)]
+    # subset genes to 10 test genes
+    # remove this when running on entire dataset
+    genes_to_test = ['ERAP2', 'LAMB1', 'POLR2J2', 'GBP7', 'GSTM3', 'CCDC163P', 'CTLA4', 'AIF1', 'BRCA1', 'IL7', 'HDHD5']
```
illusional, line 90: Should this still be here?

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -0,0 +1,449 @@
...
+
+    # Get 1Mb sliding window around each gene
+    geneloc_df = pd.read_csv(AnyPath(geneloc), sep='\t')
+    gene_ids = list(log_expression_df.columns.values)[1:]
+    geneloc_df = geneloc_df[geneloc_df.gene_name.isin(gene_ids)]
+    geneloc_df = geneloc_df.assign(left=geneloc_df.start - 1000000)
+    geneloc_df = geneloc_df.assign(right=geneloc_df.end + 1000000)
+    # subset genes to 10 test genes
+    # remove this when running on entire dataset
+    genes_to_test = ['ERAP2', 'LAMB1', 'POLR2J2', 'GBP7', 'GSTM3', 'CCDC163P', 'CTLA4', 'AIF1', 'BRCA1', 'IL7', 'HDHD5']
```
illusional, line 323: Ditto

scripts/eqtl_hail_batch/launch_generate_eqtl_spearman.py
```diff
@@ -0,0 +1,150 @@
...
+                analysis_runner_output_path = output_prefix[5:].partition('/')[-1]
+                # get access level from bucket, rather than manual input
+                access_level = bucket_name.split('-')[-1]
+                run_analysis_runner(
+                    description=f'eqtl_spearman_{cell_type}_chr{chromosome}',
+                    dataset='tob-wgs',
+                    access_level=access_level,
+                    output_dir=analysis_runner_output_path,
+                    memory='highmem',
+                    cpu=8,
```
illusional: Ditto with top comment, does this need 8cpus, or are you requiring a specific amount of memory?

scripts/eqtl_hail_batch/README.md
KatalinaBobowik: That's a good point, this might be a holdover from when we were loading in the large genotype files. I can remove it and see

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
KatalinaBobowik: This actually does stay - I'm currently only testing 10 genes, but this will change later on

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
KatalinaBobowik: as above, stays for now

scripts/eqtl_hail_batch/launch_generate_eqtl_spearman.py
KatalinaBobowik: I'll go ahead and remove these

scripts/eqtl_hail_batch/README.md
KatalinaBobowik: I've removed this now. The test seems to work fine. I think you're right in that the analysis runner is just loading files in and the batch workflow does the rest

scripts/eqtl_hail_batch/launch_conditional_analysis.py
MattWellie: Just that `exists()` can be used as a replacement for the `file_exists` method, and the `glob` can be used instead of `bucket.list_blobs(prefix=path_to_expression_files + '/', delimiter='/')`!

scripts/eqtl_hail_batch/README.md
```diff
@@ -0,0 +1,45 @@
+# Run hail batch job to generate list of eQTLs
+
+This runs a Hail batch script in order to generate a list of eQTLs from scRNA-seq expression. This code was taken from Seyhan Yazar from Joseph Powell's group at the Garvan-Weizmann Centre for Cellular Genomics, then converted into Python/hail batch. To run, use conda to install the analysis-runner. For the first round of eQTLs, execute the following command for running one cell type and chromosome (B_intermediate cells and chromosome 22 are shown here as an example):
```
MattWellie: Could you link to that original codebase here?

scripts/eqtl_hail_batch/conditional_analysis.py
```diff
@@ -0,0 +1,465 @@
...
+
+DEFAULT_DRIVER_MEMORY = '4G'
+DRIVER_IMAGE = os.getenv('CPG_DRIVER_IMAGE')
+assert DRIVER_IMAGE
+
+TOB_WGS = dataset_path('mt/v7.mt/')
+FREQ_TABLE = dataset_path('joint-calling/v7/variant_qc/frequencies.ht/', 'analysis')
+
+
+def get_number_of_scatters(residual_df, significant_snps_df):
```
MattWellie: one general note, if you include Type inference data then pylint can help catch some sneaky bugs

```suggestion
def get_number_of_scatters(residual_df: pd.DataFrame, significant_snps_df: pd.DataFrame) -> int:
```

scripts/eqtl_hail_batch/conditional_analysis.py
```diff
@@ -0,0 +1,465 @@
...
+    # Further reduce the table by only selecting SNPs needed
+    set_to_keep = hl.literal(snps_to_keep)
+    t = t.filter(set_to_keep.contains(t['snpid']))
+    # only keep SNPs where all samples have an alt_allele value
+    snps_to_remove = set(t.filter(hl.is_missing(t.n_alt_alleles)).snpid.collect())
+    if len(snps_to_remove) > 0:
+        t = t.filter(~hl.literal(snps_to_remove).contains(t.snpid))
+
+    genotype_df = t.to_pandas(flatten=True)
+    genotype_df.rename({'onek1k_id': 'sampleid'}, axis=1, inplace=True)
```
MattWellie, line 159: AFAIK you can do this (renaming the field) earlier on during the select
(added above)

scripts/eqtl_hail_batch/conditional_analysis.py
```diff
@@ -0,0 +1,465 @@
...
+    sorted_snp_positions = list(map(lambda x: x.split(':')[:2][1], sorted_snps))
+    sorted_snp_positions = [int(i) for i in sorted_snp_positions]
+    # get first and last positions, with 1 added to last position (to make it inclusive)
+    chromosome = gene_snp_test_df.snpid[0].split(':')[:1][0]
+    first_and_last_snp = chromosome + ':' + str(sorted_snp_positions[0]) + '-' + str(sorted_snp_positions[-1]+1)
+    mt = hl.filter_intervals(mt, [hl.parse_locus_interval(first_and_last_snp, reference_genome='GRCh38')])
+    t = mt.entries()
+    t = t.annotate(n_alt_alleles=t.GT.n_alt_alleles())
+    t = t.key_by(contig=t.locus.contig, position=t.locus.position)
+    t = t.select(t.alleles, t.onek1k_id, t.n_alt_alleles)
```
MattWellie:
```suggestion
    t = t.select(t.alleles, sampleid=t.onek1k_id, t.n_alt_alleles)
```

scripts/eqtl_hail_batch/conditional_analysis.py
```diff
@@ -0,0 +1,465 @@
...
+    filter_mt_job = batch.new_python_job('filter_mt')
+    copy_common_env(filter_mt_job)
+    filtered_mt_path = filter_mt_job.call(
+        prepare_genotype_info, keys_path=keys
+    )
+
+    previous_sig_snps_result = significant_snps_df  # pylint: disable=invalid-name
+    previous_residual_result = residual_df  # pylint: disable=invalid-name
+    # Perform conditional analysis for n iterations (specified in click interface)
+    for iteration in range(iterations):
```
MattWellie: You can alter the range of `range`, so you don't need to manually edit the `iterations` each time
```suggestion
    for iteration in range(2, iterations +2):
```

scripts/eqtl_hail_batch/conditional_analysis.py
```diff
@@ -0,0 +1,465 @@
...
+    t = t.select(t.alleles, t.onek1k_id, t.n_alt_alleles)
+    t = t.annotate(
+        snpid=hl.str(t.contig)
+        + ':'
+        + hl.str(t.position)
+        + ':'
+        + hl.str(t.alleles[0])
+        + ':'
+        + hl.str(t.alleles[1])
+    )
```
MattWellie: I found some syntax that will do this in-line, but I genuinely hate it
```suggestion
    t = t.annotate(snpid=hl.str(':').join(list(map(hl.str, [t.contig, t.position, t.alleles[0], t.alleles[1]]))))
```

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -0,0 +1,460 @@
...
+    gene_snp_df = snps_within_region.assign(
+        gene_id=gene_info.gene_id, gene_symbol=gene_info.gene_name
+    )
+    
+    # get genotypes from mt in order to load individual SNPs into
+    # the spearman correlation function
+    t = mt.entries()
+    t = t.annotate(n_alt_alleles=t.GT.n_alt_alleles())
+    t = t.key_by(contig=t.locus.contig, position=t.locus.position)
+    t = t.select(t.alleles, t.onek1k_id, t.n_alt_alleles)
```
MattWellie:
```suggestion
    t = t.select(t.alleles, sampleid=t.onek1k_id, t.n_alt_alleles)
```

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -0,0 +1,460 @@
...
+    set_to_keep = hl.literal(snps_to_keep)
+    t = t.filter(set_to_keep.contains(t['snpid']))
+    # only keep SNPs where all samples have an alt_allele value
+    # (argument must be a string, a bytes-like object or a number)
+    snps_to_remove = set(t.filter(hl.is_missing(t.n_alt_alleles)).snpid.collect())
+    if len(snps_to_remove) > 0:
+        t = t.filter(~hl.literal(snps_to_remove).contains(t.snpid))
+    
+    genotype_df = t.to_pandas(flatten=True)
+    genotype_df.rename({'onek1k_id': 'sampleid'}, axis=1, inplace=True)
```
MattWellie: paired with above `select` statement
```suggestion
```

MattWellie (commented):
This is way outside of my comfort zone, but looks pretty good to me! I'd be happy to approve, but would rather delegate to someone who knows the area better :s

scripts/eqtl_hail_batch/README.md
KatalinaBobowik: Great idea! The publicly-available code wasn't available when I when I was writing this, but now that the OneK1K paper is out, it is. I'll add that link in :)

scripts/eqtl_hail_batch/conditional_analysis.py
KatalinaBobowik: That's a great tip! (I saw this in your and some other people's code earlier, and wasn't sure what exactly it did.) Thank you

scripts/eqtl_hail_batch/conditional_analysis.py
KatalinaBobowik: great tip, thanks! Will add that in and update the rest

scripts/eqtl_hail_batch/conditional_analysis.py
KatalinaBobowik: 🤩 👌

scripts/eqtl_hail_batch/launch_conditional_analysis.py
```diff
@@ -0,0 +1,160 @@
...
+                        *('--residuals', residuals),
+                        *('--significant-snps', significant_snps),
+                        *('--output-prefix', output_prefix),
+                        *(
+                            ['--test-subset-genes', str(test_subset_genes)]
+                            if test_subset_genes
+                            else []
+                        ),
+                        *('--keys', keys),
+                    ],
```
illusional, line 153:
```suggestion
                    ],
                    image="australia-southeast1-docker.pkg.dev/cpg-common/images/multipy:0.16",
```

scripts/eqtl_hail_batch/README.md
```diff
@@ -0,0 +1,45 @@
+# Run hail batch job to generate list of eQTLs
+
+This runs a Hail batch script in order to generate a list of eQTLs from scRNA-seq expression. This [code](https://github.com/powellgenomicslab/onek1k_phase1/tree/main/single_cell_cis_eQTL_mapping) was taken from Seyhan Yazar from Joseph Powell's group at the Garvan-Weizmann Centre for Cellular Genomics, then converted into Python/hail batch. To run, use conda to install the analysis-runner. For the first round of eQTLs, execute the following command for running one cell type and chromosome (B_intermediate cells and chromosome 22 are shown here as an example):
+
+```sh
+analysis-runner --dataset tob-wgs \
+    --access-level test --output-dir "plasma/chr22/v0" \
+    --description "eqtl batch job" \
+    python3 generate_eqtl_spearman.py \
```
illusional, line 35:
```suggestion
    --image australia-southeast1-docker.pkg.dev/cpg-common/images/multipy:0.16 \
    python3 generate_eqtl_spearman.py \
```

scripts/eqtl_hail_batch/conditional_analysis.py
```diff
@@ -0,0 +1,458 @@
...
+    copy_common_env,
+    init_batch,
+    remote_tmpdir,
+)
+from cloudpathlib import AnyPath
+import click
+
+
+DEFAULT_DRIVER_MEMORY = '4G'
+DRIVER_IMAGE = os.getenv('CPG_DRIVER_IMAGE')
```
illusional:
```suggestion
DRIVER_IMAGE = "australia-southeast1-docker.pkg.dev/cpg-common/images/multipy:0.16"
```

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -0,0 +1,458 @@
...
+DEFAULT_DRIVER_MEMORY = '4G'
+DRIVER_IMAGE = 'australia-southeast1-docker.pkg.dev/cpg-common/images/multipy:0.16'
+assert DRIVER_IMAGE
+
+TOB_WGS = dataset_path('mt/v7.mt/')
+FREQ_TABLE = dataset_path('joint-calling/v7/variant_qc/frequencies.ht/', 'analysis')
+
+
+def filter_lowly_expressed_genes(expression_df):
+    """Remove genes with low expression in all samples"""
```
katiedelange: It would be great if you could expand this comment (and the equivalent one for the other functions) to define what you expect expression_df to look like - e.g. 

"""Remove genes with low expression in all samples

Args:
expression_df (matrix): a matrix with samples as rows and genes as columns, containing expression values as transcripts per million (TPM)

Returns:
matrix: a filtered version of the input matrix, after removing columns (genes) with 0 values in more than 10% of the rows (samples)
"""

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -0,0 +1,458 @@
...
+def remove_sc_outliers(df):
+    # remove outlier samples, as identified by sc analysis
+    outliers = ['966_967', '88_88']
+    df = df[-df.sampleid.isin(outliers)]
+
+    return df
+
+
+def get_number_of_scatters(expression_df, geneloc_df):
+    """get index of total number of genes"""
```
katiedelange: Similarly, can you explicitly define geneloc_df, and what format it should take? While it can be inferred from the code, it's nice to have it made obvious upfront.

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -0,0 +1,458 @@
...
+
+    """get logged expression values"""
+
+    expression_df = filter_lowly_expressed_genes(expression_df)
+    sample_ids = expression_df.iloc[:, 0]
+    to_log = expression_df.iloc[:, 1:].columns
+    log_expression_df = expression_df[to_log].applymap(lambda x: np.log(x + 1))
+    # subset genes to 10 test genes
+    # remove this when running on entire dataset
+    genes_to_test = ['ERAP2', 'LAMB1', 'POLR2J2', 'GBP7', 'GSTM3', 'CCDC163P', 'CTLA4', 'AIF1', 'BRCA1', 'IL7', 'HDHD5']
```
katiedelange, line 115: This might actually be some nice functionality to include permanently - I can imagine some use cases where someone might want to come back and run just a very specific gene or set of genes of interest with new parameters. Perhaps you could add genes_to_test as an optional parameter to the relevant functions (default = null), and then a short if condition to subset the expression_df at the outset if a gene set is provided, or leave it unchanged if gene set is null.

Just a suggestion and by no means something you need to implement right now.

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -0,0 +1,458 @@
...
+def calculate_log_cpm(expression_df, output_prefix):
+    """Calculate log cpm for each cell type and chromosome"""
+
+    expression_df = filter_lowly_expressed_genes(expression_df)
+    sample_ids = expression_df.iloc[:, 0]
+    # remove sampleid column and get log expression
+    # this can only be done on integers
+    expression_df = expression_df.iloc[:, 1:]
+    cpm_df = expression_df.apply(lambda x: (x / sum(x)) * 1000000, axis=0)
+    log_cpm = np.log(cpm_df + 1)
```
katiedelange, line 141: Just checking I understand why the log(x **+ 1**) transformation is used (I see it's in Seyhan's script too) - this is just to deal with the fact that there are zero values in the expression data?

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -0,0 +1,458 @@
...
+        # VQSR does not filter out low quality genotypes. Filter these out
+        mt = mt.filter_entries(mt.GQ <= 20, keep=False)
+        # filter out samples with a genotype call rate > 0.8 (as in the gnomAD supplementary paper)
+        n_samples = mt.count_cols()
+        call_rate = 0.8
+        mt = mt.filter_rows(
+            hl.agg.sum(hl.is_missing(mt.GT)) > (n_samples * call_rate), keep=False
+        )
+        # filter out variants with MAF < 0.01
+        ht = hl.read_table(FREQ_TABLE)
```
katiedelange, line 212: So here we use the frequencies that were calculated _before_ filtering out samples with low call rates etc. I don't think this is a problem (I assume the extra time required to recalculate the frequencies each round is not insignificant?), but just to note that this means we are likely to have a few variants that actually have MAF < 0.01 in our final dataset in case it ever comes up in a plot and confuses us.

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -0,0 +1,458 @@
...
+        # only keep samples that have rna-seq expression data
+        samples_to_keep = set(log_expression_df.sampleid)
+        set_to_keep = hl.literal(samples_to_keep)
+        mt = mt.filter_cols(set_to_keep.contains(mt['onek1k_id']))
+        mt.write(filtered_mt_path)
+
+    return filtered_mt_path
+
+
+def calculate_residuals(expression_df, covariate_df, output_prefix):
```
katiedelange, line 238: Ditto the points from above: a little more information on the inputs/outputs and their required format would be great.

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -0,0 +1,458 @@
...
+@click.option(
+    '--geneloc', required=True, help='A TSV of start and end positions for each gene'
+)
+@click.option(
+    '--covariates', required=True, help='A TSV of covariates to calculate residuals'
+)
+@click.option(
+    '--keys',
+    required=True,
+    help='A TSV of sample ids to convert external to internal IDs',
```
katiedelange: For each of these help statements, a little more detail on the expected file format (expected columns, units used for expression, what should be rows and what should be columns) would be useful.

scripts/eqtl_hail_batch/launch_conditional_analysis.py
```diff
@@ -0,0 +1,161 @@
...
+import os
+import click
+
+from analysis_runner.cli_analysisrunner import run_analysis_runner
+from google.cloud import storage
+
+
+# def run_analysis_runner(*args, **kwargs):
+#     """Run analysis runner for all cell types and chromosomes"""
+#     print(args, kwargs)
```
katiedelange: Should this commented out code still be here?

scripts/eqtl_hail_batch/launch_conditional_analysis.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+@click.command()
+@click.option(
+    '--cell-types',
+    default=None,
+    help='List of cell types to test',
```
katiedelange: Information on the available cell types and how to specify them would be helpful. Are these determined by e.g. the column names in a certain input file?

scripts/eqtl_hail_batch/launch_conditional_analysis.py
```diff
@@ -0,0 +1,161 @@
...
+)
+@click.option(
+    '--chromosomes',
+    required=True,
+    help='List of chromosome numbers to run eQTL analysis on. Space separated, as one argument',  # noqa: E501; pylint: disable=line-too-long
```
katiedelange: (Nitpick) This could be split over two lines using the construct you use on line 48, so you don't have to disable the linter.

scripts/eqtl_hail_batch/launch_generate_eqtl_spearman.py
```diff
@@ -0,0 +1,149 @@
...
+# def run_analysis_runner(*args, **kwargs):
+#     """Run analysis runner for all cell types and chromosomes"""
+#     print(args, kwargs)
+
+
+@click.command()
+@click.option(
+    '--cell-types',
+    default=None,
+    help='List of cell types to test',
```
katiedelange: Ditto on knowing what cell types can be provided here.

scripts/eqtl_hail_batch/conditional_analysis.py
```diff
@@ -0,0 +1,459 @@
...
+assert DRIVER_IMAGE
+
+TOB_WGS = dataset_path('mt/v7.mt/')
+FREQ_TABLE = dataset_path('joint-calling/v7/variant_qc/frequencies.ht/', 'analysis')
+
+
+def get_number_of_scatters(residual_df: pd.DataFrame, significant_snps_df: pd.DataFrame) -> int:
+    """get index of total number of genes"""
+
+    # Identify the top eSNP for each eGene and assign remaining to df
```
katiedelange: (Nitpick): you don't actually assign the remaining SNPs to the df in this function/chunk of code (I think?)

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -0,0 +1,458 @@
...
+        # filter out samples with a genotype call rate > 0.8 (as in the gnomAD supplementary paper)
+        n_samples = mt.count_cols()
+        call_rate = 0.8
+        mt = mt.filter_rows(
+            hl.agg.sum(hl.is_missing(mt.GT)) > (n_samples * call_rate), keep=False
+        )
+        # filter out variants with MAF < 0.01
+        ht = hl.read_table(FREQ_TABLE)
+        mt = mt.annotate_rows(freq=ht[mt.row_key].freq)
+        mt = mt.filter_rows(mt.freq.AF[1] > 0.01)
```
katiedelange, line 214: Should this perhaps be >= 0.01 - here we keep anything with a MAF > 0.01, and dropping everything MAF <= 0.01?

scripts/eqtl_hail_batch/conditional_analysis.py
```diff
@@ -0,0 +1,459 @@
...
+        # filter out samples with a genotype call rate > 0.8 (as in the gnomAD supplementary paper)
+        n_samples = mt.count_cols()
+        call_rate = 0.8
+        mt = mt.filter_rows(
+            hl.agg.sum(hl.is_missing(mt.GT)) > (n_samples * call_rate), keep=False
+        )
+        # filter out variants with MAF < 0.01
+        ht = hl.read_table(FREQ_TABLE)
+        mt = mt.annotate_rows(freq=ht[mt.row_key].freq)
+        mt = mt.filter_rows(mt.freq.AF[1] > 0.01)
```
katiedelange, line 91: Should this perhaps be >= 0.01 - here we keep anything with a MAF > 0.01, and dropping everything MAF <= 0.01?

scripts/eqtl_hail_batch/conditional_analysis.py
```diff
@@ -0,0 +1,459 @@
...
+    # save sampleids before filtering redidual_df
+    sample_ids = residual_df.sampleid
+    residual_df = residual_df[gene_ids]
+    # reassign sample ids
+    residual_df = residual_df.assign(sampleid=sample_ids.to_list())
+
+    # Subset genotype file for the significant SNPs
+    genotype_df = get_genotype_df(
+        filtered_mt_path, residual_df, esnp1
+    )
```
katiedelange, line 205: I'm somewhat struggling to follow what's happening here - sorry, I think my python is too rusty! But I can't find when the lead SNP is actually being adjusted and how..

So is `esnp1` now all significant variants in the locus _minus_ the top SNP (did `.reset_index()` on line 142 drop it?)? And then we're now fetching the genotypes for all those variants in `esnp1`?

But if so, later on line 164 how are we managing to retrieve the ID of the top SNP from `esnp1` and `genotype_df`?

katiedelange (commented):
Nice work! Overall this looks really great and was easy to read. 

The biggest issue is that I'm a little confused on how the conditional analysis is working - it's likely that this is just a lack of understanding on my end, but it would be great if you could help me see where the lead SNP adjustments are actually happening.

Other than that, there are just a few minor requests around expanding some of the documentation to make it easier to follow the code, and some clarification around the MAF filter (< 0.01, or <= 0.01).

scripts/eqtl_hail_batch/conditional_analysis.py
KatalinaBobowik: Sorry, I'm a little confused- when you say "dropping everything MAF <= 0.01", isn't that what it's currently doing?

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
KatalinaBobowik: as below

scripts/eqtl_hail_batch/conditional_analysis.py
KatalinaBobowik: Good catch! I removed that bit

scripts/eqtl_hail_batch/launch_conditional_analysis.py
KatalinaBobowik: that was used for a dry run, but agree should be removed now.

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
KatalinaBobowik: great idea, I'll expand that for all the functions to make it easier to understand input --> output

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
KatalinaBobowik: no problem, I'll add this in

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
KatalinaBobowik: My understanding is that the log transformation is used in Seyhan's code to remove skewness/normalise the data. However, here I'm transforming counts to logCPM, which was something Daniel E, Daniel R, and I discussed for incorporating into the violin plots. 

On another note, after looking into the data types (addressing your comment above), this data does not look like count data (which is what this function expects). From speaking with Seyhan earlier, I was under the assumption this was count data, so have to dig into this a bit deeper (I've sent her a slack message so should find out soon).

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
KatalinaBobowik: I can look into this, too

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
KatalinaBobowik: That's a good point! I'll make a note of that, so that it doesn't confuse use later.

scripts/eqtl_hail_batch/conditional_analysis.py
KatalinaBobowik: esnp1 is the most significant SNP (reset index simply moves the gene_name, which would otherwise be the index, to within the df). get_genotype_df() then gets the genotypes for all samples, exclusively for that SNP. A linear regression model is then fit to the data (residual expression values for each sample x genotype), and the residuals of that are taken

scripts/eqtl_hail_batch/launch_conditional_analysis.py
KatalinaBobowik: At the moment, there are nearly 30 cell types, so a bit tricky to list out all of them. I expanded the help statement to point to the available list of cell types in `gs://cpg-tob-wgs-main/scrna-seq/grch38_association_files/expression_files/`

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
KatalinaBobowik: Ok, just to confirm from Seyhan, values represent normalised counts, where in each celltype file, matrix values are the average number of molecules for each gene detected in each person. These, in essence, represent a pseudocount matrix, and in theory, are ok to use for the logCPM calculation

KatalinaBobowik:
@katiedelange apart from including some permanent code that incorporates a specific gene set to filter on, these suggestions should all now be addressed

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
katiedelange: Great, sounds good - thanks for clarifying that.

scripts/eqtl_hail_batch/conditional_analysis.py
katiedelange: Sorry, that was a _terribly_ written comment from me! Yes, filtering out MAF <= 0.01 is what the code is currently doing - but your comment says we're filtering out MAF < 0.01.

So this is being super nit picky, but just to make sure the description of what we do matches up with what we actually do (and what Seyhan did) so that it's easy to write up an accurate methods description later.

scripts/eqtl_hail_batch/conditional_analysis.py
katiedelange: Hmm, ok, I think I see - it had confused me because the original script [pulled all genotypes](https://github.com/powellgenomicslab/onek1k_phase1/blob/main/single_cell_cis_eQTL_mapping/round2.run_spearman_rank_test.R#L103), but I see now that that is unnecessary. So am I interpreting it right that the process is basically this?

**Base analysis (Round 1)**
- _Input: all SNP genotypes, full scRNA-seq expression data_
- **Step 1**: fit a model to expression data to account for experimental covariates (sex, age, PCAs 1-4, Peer Factors 1 & 2), take residuals (= adjusted expression)
- **Step 2**: test association of all SNPs to adjusted expression
- _Output: set of significantly associated SNPs, expression data adjusted for covariates_

**Conditional analysis (Round 2 +)**
- _Input: significant SNP genotypes from previous round, expression data adjusted for experimental covariates (or in round 3+, adjusted for experimental covariates and previous round lead SNPs)_
- **Step 1**: fit a model to adjusted expression data to account for genotype of most significant SNP in locus, take residuals (= SNP & covariate adjusted expression)
- **Step 2**: test association of significant SNPs from previous round against SNP & covariate adjusted expression data 
- _Output: set of significantly associated SNPs, expression data adjusted for lead SNP & covariates_

katiedelange (commented):
Thanks Kat, the additional documentation is really helpful! I think I now understand the logic behind the process being taken here, and I'll mark this for approval as soon as you confirm if we are actually on the same page or not (see comment in the conditional analysis script).

scripts/eqtl_hail_batch/conditional_analysis.py
KatalinaBobowik: yes, that's exactly right! Would it be helpful to have this somewhere in the README so that it's more clear?

scripts/eqtl_hail_batch/conditional_analysis.py
KatalinaBobowik: makes sense. Comments in both scripts should now reflect that I'm filtering in this way (so there's no confusion later)

scripts/eqtl_hail_batch/conditional_analysis.py
katiedelange: Thanks! Not a blocker on this PR, but if you have time to update the README to explain the process then I at least would find that helpful when coming back to this months later :)

katiedelange (approved):
Looks good to me, happy for this to go ahead.

scripts/eqtl_hail_batch/conditional_analysis.py
KatalinaBobowik: Fantastic, I can add that in there.

KatalinaBobowik:
> Looks good to me, happy for this to go ahead.

Thanks @katiedelange !

katiedelange (approved):

scripts/eqtl_hail_batch/conditional_analysis.py
```diff
@@ -0,0 +1,528 @@
...
+from cpg_utils.hail_batch import (
+    dataset_path,
+    output_path,
+    copy_common_env,
+    init_batch,
+    remote_tmpdir,
+)
+from cloudpathlib import AnyPath
+import click
+from multipy.fdr import qvalue
```
illusional: Don't import this at the top level

scripts/eqtl_hail_batch/conditional_analysis.py
```diff
@@ -0,0 +1,528 @@
...
+
+    return significant_snps_df
+
+
+def merge_significant_snps_dfs(*df_list):
+    """
+    Merge list of list of sig_snps dataframes
+    """
+
+    merged_sig_snps = pd.concat(df_list)
```
illusional, line 378: Import multipy here instead (this will give you a pylint error which you can suppress) 

```suggestion
    from multipy.fdr import qvalue 
    merged_sig_snps = pd.concat(df_list)
```

scripts/eqtl_hail_batch/conditional_analysis.py
```diff
@@ -0,0 +1,528 @@
...
+                iteration,
+                gene_idx,
+                previous_residual_result,
+                previous_sig_snps_result,
+                filtered_mt_path
+            )
+            sig_snps_dfs.append(gene_result)
+
+        merge_job = batch.new_python_job(name='merge_scatters')
+        merge_job.cpu(2)
```
illusional, line 506:
```suggestion
        merge_job.image(MULTIPY_IMAGE)
        merge_job.cpu(2)
```

scripts/eqtl_hail_batch/conditional_analysis.py
```diff
@@ -0,0 +1,528 @@
...
+    init_batch,
+    remote_tmpdir,
+)
+from cloudpathlib import AnyPath
+import click
+from multipy.fdr import qvalue
+
+
+DEFAULT_DRIVER_MEMORY = '4G'
+DRIVER_IMAGE = 'australia-southeast1-docker.pkg.dev/cpg-common/images/multipy:0.16'
```
illusional:
```suggestion
MULTIPY_IMAGE = 'australia-southeast1-docker.pkg.dev/cpg-common/images/multipy:0.16'
DRIVER_IMAGE = # get equivalent get driver image
```

scripts/eqtl_hail_batch/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Running via the analysis-runner:
+
+To run, use conda to install the analysis-runner. For the first round of eQTLs, execute the following command for running one cell type and chromosome (B_intermediate cells and chromosome 22 are shown here as an example):
+
+```sh
+analysis-runner --dataset tob-wgs \
+    --access-level test --output-dir "plasma/chr22/v0" \
+    --description "eqtl batch job" \
+    --image australia-southeast1-docker.pkg.dev/cpg-common/images/multipy:0.16 \
```
illusional: Sorry to backtrack, but you can remove this line now (to get the regular default driver imagE)

illusional (changes requested):
Requesting changes to multipy to make more solid the changes to driver image

scripts/eqtl_hail_batch/launch_conditional_analysis.py
MattWellie: I just came back to this because I needed it as a code example, and realised what a pointless response this was ^^

Advantages with cloudpathlib are that `CloudPaths` or `AnyPaths` are cloud-provider agnostic. The library is built to recognise what a GCP/AWS/Azure path looks like, and will handle everything appropriately behind the scenes. `AnyPath` also allows a mix of local and cloud paths, and it will work out how to access the file.

On one hand it's cleaner code, because you don't have to open a bucket/client connection etc. etc., but it's also potentially better code, as we can switch to any other cloud provider and no code changes are required.

vladsavelyev (approved):
Just adding another green tick :)

scripts/eqtl_hail_batch/conditional_analysis.py
```diff
@@ -16,14 +16,14 @@
...
 from cloudpathlib import AnyPath
 import click
-from multipy.fdr import qvalue
 
 
 DEFAULT_DRIVER_MEMORY = '4G'
-DRIVER_IMAGE = 'australia-southeast1-docker.pkg.dev/cpg-common/images/multipy:0.16'
-assert DRIVER_IMAGE
+MULTIPY_IMAGE = 'australia-southeast1-docker.pkg.dev/cpg-common/images/multipy:0.16'
+assert MULTIPY_IMAGE
```
illusional: Nitpick: Can remove this as its set to a defined value

```suggestion
```

scripts/eqtl_hail_batch/conditional_analysis.py
```diff
@@ -375,6 +372,8 @@ def merge_significant_snps_dfs(*df_list):
     Merge list of list of sig_snps dataframes
     """
 
+    from multipy.fdr import qvalue
```
illusional, line 376: Maybe just add a comment here, something like:

```suggestion
    # This PythonJob is run in the multipy container, 
    # do the import here so it's not run in the driver container
    from multipy.fdr import qvalue 
```

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -416,20 +439,27 @@ def spearman_correlation(df):
...
-    spearman_df.snp_id = spearman_df.apply(lambda x: ':'.join(x[['chrom', 'bp', 'a1', 'a2']]), axis=1)
+    spearman_df['association_id'] = spearman_df.apply(lambda x: ':'.join(x[['chrom', 'bp', 'a1', 'a2', 'gene_symbol', 'cell_type_id', 'round']]), axis=1)
+    spearman_df['variant_id'] = spearman_df.apply(lambda x: ':'.join(x[['chrom', 'bp', 'a2']]), axis=1)
+    spearman_df['snp_id'] = spearman_df.apply(lambda x: ':'.join(x[['chrom', 'bp', 'a1', 'a2']]), axis=1)
+    
     return spearman_df
 
 
 def merge_df_and_convert_to_string(*df_list):
     """Merge all Spearman dfs and convert to string using .to_string() on df"""
```
illusional, line 458: Can you include more information on what qvalue is doing in this function, maybe even update the function name?

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -536,6 +567,7 @@ def main(
 
     merge_job = batch.new_python_job(name='merge_scatters')
     merge_job.cpu(2)
+    merge_job.image(MULTIPY_IMAGE)
```
illusional, line 573: Small comment would be good for why using the multipy image here

illusional (approved):
A few nitpicks, mostly around comments on where / why to use the multipy image instead of the regular driver.

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
KatalinaBobowik: added in a comment to explain this a bit more

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
KatalinaBobowik: added

PR #98, opened 2022-07-12
Update browser output for front end pipeline

### LINKED ISSUE #97: Update eqtl pipeline outputs for browser
# Pipeline output specification

This issue outlines preferred data formats of the analysis pipeline's output.

## Associations
Updates to the columns currently output by the eqtl pipeline.

| Name              | Format                                            | 
| ----------------- | ------------------------------------------------- |
| chrom             | string without 'chr'                              |
| association_id    | chrom:bp:a1:a2:gene_id:cell_type_id:round         |
| cell_type_id      | string                                            |
| cell_type_name    | long form description of cell type (if available) |
| variant_id        | chrom:bp:a2                                       |
| snp_id            | chrom:bp:a1:a2                                    |
| locus             | chrom:bp                                          |
| beta              | number                                            |
| functional_annotation | string from VEP               |
| is_esnp | boolean indicating if eQTL was used to condition the next round

Example:

| association_id | cell_type_id | cell_type_name | chrom | variant_id | snp_id | beta | functional_annotation |
| -------------- | ------------ | -------------- | ----- | ---------- | ------ | ---- | --------------- |
| 6:49458097:A:G:ENSG00000031691:bmem:1 | bmem | Memory B Cell | 6 | 6:49458097:G | 6:49458097:A:G | 0.7 | missense_variant |

The field `id` should be renamed to `association_id` and `snpid` renamed to `snp_id` due to the data format expected by the web UI.

Outputs files in the form `chr1.tsv.gz` or `chr1.parquet` where each file combines results from all cell types and conditioning rounds. If saving as a parquet file, the file should be indexed on `global_bp` with a step size of 2 megabases (MB).

## Gene expression distribution (logCPM)
A table mapping gene identifier and cell type id to a data structure summarising gene expression information from samples grouped by cell type id.

Example:

| gene_id | gene_symbol | cell_type_id | data |
| ------- | ----------- | ------------ | ---- |
| ENSG00000134202 | GSTM3 | bmem | \<structure> |

The `data` field has the dictionary-like form:

```
{
    bin_counts: number[],
    bin_edges: number[],
    n_samples: number,
    min: number,
    max: number,
    mean: number,
    median: number,
    q1: number,
    q3: number,
    iqr: number,
    iqr_min: number, // q1 - 1.5 * (q3 - q1)
    iqr_max: number, // q3 + 1.5 * (q3 - q1)
}
```

Outputs a single file named `gene_expression.parquet`.

## Association effect on gene expression (logCPM)
A table mapping eQTLs to a data structure summarising gene expression information. Samples are grouped based on genotype (0 - hom ref, 1 - het alt, 2 - hom alt) and cell type id. Gene expression summary information is then computed for each grouping.

Example:

| snp_id | global_bp | gene_id | gene_symbol | cell_type_id | genotype | data |
| ------ | --------- | ------- | ----------- | ------------ | -------- | ---- |
| 1:108735360:G:C | 108735360 | ENSG00000134202 | GSTM3 | bmem | 0 | \<structure> |
| 1:108735360:G:C | 108735360 | ENSG00000134202 | GSTM3 | bmem | 1 | \<structure> |
| 1:108735360:G:C | 108735360 | ENSG00000134202 | GSTM3 | bmem | 2 | \<structure> |

The `genotype` field indicates the number of `alt (a2)` alleles in the sample. The field `global_bp` is required to index the table to reduce query costs while increasing query speed. The `snp_id` field should match that in the association table output. The `data` field has the dictionary-like form (same as above):

```
{
    bin_counts: number[],
    bin_edges: number[],
    n_samples: number,
    min: number,
    max: number,
    mean: number,
    median: number,
    q1: number,
    q3: number,
    iqr: number,
    iqr_min: number, // q1 - 1.5 * (q3 - q1)
    iqr_max: number, // q3 + 1.5 * (q3 - q1)
}
```

Outputs a single file named `eqtl_effect.parquet` indexed on `global_bp` with a step size of 2 megabases (MB). This table will be very large on the full data set, so you could first try computing only for associations with an `FDR < 0.05`

## LD information
A table mapping an locus identifier to a structure of loci in LD within **1MB(?)** flank either side.

Example:

| locus | global_bp | data |
| ----- | --------- | ---- |
| 1:108735360 | 108735360 | \<structure> | 

The field `global_bp` is required to index the table to reduce query costs while increasing query speed. The `data` field has the dictionary-like form mapping a locus to a dictionary keyed by locus identifier where values are LD scores:

```
{
   1:108635360: 0.1,
   ...,
   1:108735360: 1.0,
   1:108735361: 0.9,
   1:108735362: 0.1,
   ... 
   1:109735360: 0.5,
}
```

Outputs a single file named `ld.parquet` indexed on `global_bp` with a step size of 2 megabases (or double the size flank used to compute LD)

## Cell type metadata
A table with cell type details:

Example:

| cell_type_id | cell_type_name | description |
| ------- | ----------- | ------------ |
| bmem | Memory B Cell | TCL1A- FCER2- B cell |

Outputs a single file named `cell_types.tsv`.

# Tasks
- [x] Association file update
- [x] Gene expression stats
- [x] Association effect stats
- [ ] LD information
- [x] Cell type metadata

KatalinaBobowik:
These updates were made to address the GitHub issues outlined [here](https://github.com/populationgenomics/tob-wgs/issues/97). The association output has been modified so that it is easier to read into the front-end workflow, as well as reduce computational cost by combining into one script. *Note some code may change once Hail Batch is updated tomorrow morning (see [this post](https://centrepopgen.slack.com/archives/C018KFBCR1C/p1657589184890679)). 

@daniaki and @danielreti , I've addressed the changes in the doc, but please let me know if there's something that should be changed or added. I also haven't added in the `LD information` section, as this will be run on the unique union set of significant SNPs across all cell types after running the association analysis on the whole genome.


scripts/eqtl_hail_batch/calculate_ld.py
```diff
@@ -0,0 +1,64 @@
...
+    # if running on entire genome, concatenate all significant
+    # correlation matrices together, then read in as one df
+    # example: https://stackoverflow.com/questions/20906474/import-multiple-csv-files-into-pandas-and-concatenate-into-one-dataframe
+    significant_snps_path = f'{input_path}correlation_results.csv'
+    significant_snps_df = pd.read_csv(
+        significant_snps_path, sep=' ', skipinitialspace=True
+    )
+    t = hl.Table.from_pandas(significant_snps_df)
+    # only keep rows whose FDR is < 0.05
+    # t = t.filter(t.fdr < 0.05)
```
katiedelange: Should this be uncommented? Alternatively, if it's no longer needed because you only load significant SNPs into the dataframe above, can you expand the significant SNPs comment to include this threshold information (FDR < 0.05) before removing the unnecessary lines.

scripts/eqtl_hail_batch/calculate_ld.py
```diff
@@ -0,0 +1,64 @@
...
+    mt = mt.add_row_index()
+    # turn matrix into table and save, in order to reference row idx
+    mt_path = f'{input_path}significant_snps.mt'
+    mt.write(mt_path)
+    # perform ld calculation
+    ld = hl.ld_matrix(mt.GT.n_alt_alleles(), mt.locus, radius=2e6)
+    # only calculate the upper triangle 
+    ld = ld.sparsify_triangle()
+    table = ld.entries()
+    # filter out entries with an LD score less than 0.2
```
katiedelange: Given we are only dealing with significant SNPs within small loci, is this filtering necessary to maintain reasonable table sizes, or could we drop it? Two significant SNPs with such low LD between them might occasionally be interesting (depending on the shape of the association signal in the locus, where these variants sit, and the relationship between them) and it would be nice to be able to display the actual LD between them if possible.

If we need to keep the filtering, then we at least need to be clear in the browser when no value = very low LD, versus no value = not tested.

scripts/eqtl_hail_batch/calculate_ld.py
```diff
@@ -0,0 +1,64 @@
...
+    print(f'Printing table: {t.show()}')
+    t = t.key_by('global_bp')
+    # filter mt to positions which are in significant_snps table
+    mt = mt.filter_rows(hl.is_defined(t[mt.global_bp]))
+    # add row index to be able to remap
+    mt = mt.add_row_index()
+    # turn matrix into table and save, in order to reference row idx
+    mt_path = f'{input_path}significant_snps.mt'
+    mt.write(mt_path)
+    # perform ld calculation
```
katiedelange: Nitpick: can you expand the comment to explain that we are calculating LD for a 2MB window - radius=2e-6 is a bit of a confusing parameter name.

scripts/eqtl_hail_batch/conditional_analysis.py
```diff
@@ -134,29 +134,34 @@ def get_genotype_df(filtered_mt_path, residual_df, gene_snp_test_df):
...
     # only keep SNPs where all samples have an alt_allele value
-    snps_to_remove = set(t.filter(hl.is_missing(t.n_alt_alleles)).snpid.collect())
+    snps_to_remove = set(t.filter(hl.is_missing(t.n_alt_alleles)).snp_id.collect())
     if len(snps_to_remove) > 0:
-        t = t.filter(~hl.literal(snps_to_remove).contains(t.snpid))
+        t = t.filter(~hl.literal(snps_to_remove).contains(t.snp_id))
 
     genotype_df = t.to_pandas(flatten=True)
     genotype_df.rename({'onek1k_id': 'sampleid'}, axis=1, inplace=True)
+    # chromosome must be turned into a number solely
```
katiedelange: Is this change just for cosmetics when displaying, or because we are storing chromosome as a numeric column to make the database more efficient etc? If we want to treat things as numeric, there may need to be some additional transformation here to handle text chromosomes (X, Y, MT).

Note that using numeric encoding also opens a can of worms, as there aren't consistent maps for these chromosomes to numbers - i.e. Plink uses X=23, Y=24, XY=25, MT=26 vs most other software uses X=23, Y=24, MT=25. So we need to make an explicit decision on these encodings and ensure they are used consistently everywhere.

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -195,7 +202,7 @@ def prepare_genotype_info(keys_path, expression_path):
         expression_df = pd.read_csv(AnyPath(expression_path), sep='\t')
         log_expression_df = get_log_expression(expression_df)
         mt = hl.read_matrix_table(TOB_WGS)
-        mt = hl.experimental.densify(mt)
+        # mt = hl.experimental.densify(mt)
```
katiedelange: Can this be restored now, following [this fix](https://github.com/populationgenomics/hail/pull/206)?

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -388,6 +404,62 @@ def run_spearman_correlation_scatter(
...
+
+        def create_struct(gene, group):
+            hist, bin_edges = np.histogram(group[gene], bins=10)
+            n_samples = group[gene].count()
+            min_val = group[gene].min()
+            max_val = group[gene].max()
+            mean_val = group[gene].mean()
+            q1, median_val, q3 = group[gene].quantile([0.25, 0.5, 0.75])
+            iqr = q3 - q1
+            # not sure what this value is doing here
```
katiedelange: What is this comment referring to - if the interquartile ranges aren't used for the violin plots, do we need to calculate them and include them in the struct?

daniaki: We were planning to use `iqr_min` and `iqr_max` to draw boxplot whiskers, but I've just realised that whiskers are drawn up/down to the largest/smallest value within `iqr_max/iqr_min` and not those values themself. Is it ok to omit the calculation of `iqr_min/iqr_max` and use the `min/max` as the whisker end points instead?

scripts/eqtl_hail_batch/conditional_analysis.py
KatalinaBobowik: This change is to address the specifications outlined in the [GitHub issues doc](https://github.com/populationgenomics/tob-wgs/issues/97). @daniaki and @danielreti might know best about why the change is preferred, and if it's not, I can change it back (it would actually make certain parts of my script cleaner, since hail needs the 'chr' prefix to designate it's GRCh38). However, in saying that, everything here is a string representation of the number, so non-numeric.

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
KatalinaBobowik: yes, sorry that as well is a holdover. I have now uncommented this

scripts/eqtl_hail_batch/calculate_ld.py
KatalinaBobowik: updated :)

scripts/eqtl_hail_batch/calculate_ld.py
KatalinaBobowik: I'm happy to remove it, only previously the majority of rows were taken up by very low LD values. in GTEx, the minimum they use is 0.1, but in a data meeting I think we settled on 0.2. Very happy to change this though and not use any filter, as, like you said, we're only dealing with significant SNPs now. Regardless, if we set some minimum value, I agree that we should make the clear in the browser. 
Just a side note, the `calculate_ld.py` script will need further refining as I test it once all the significant SNPs come in, so this can be a conversation we have with the Daniels, or I can test and see how many results I actually get and how large the output matrix is. @daniaki or @danielreti , do you have any thoughts?

scripts/eqtl_hail_batch/calculate_ld.py
KatalinaBobowik: You're 100% correct- this should be uncommented (I think that's a holdover from testing).

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
KatalinaBobowik: Sorry, that's a very disconcerting comment to have in code 🤦‍♀️ As @daniaki said, that's a holdover from previously, when I wasn't sure what the calculation was used for, as I got it from the GitHub issues doc. What I found out is that that calculation is used to identify outliers (hence, the 1.5 constant). @daniaki no problem, so since the min and max value are already there, just keep delete the `iqr_min` and `iqr_max`? I'll go ahead and remove them from the code and the data struct

scripts/eqtl_hail_batch/conditional_analysis.py
daniaki: This format was to match the output I got from Joseph's group, but it shouldn't cause any major issues on our end if you change it back. Feel free to change it back particularly if it makes other parts of the script cleaner.

scripts/eqtl_hail_batch/calculate_ld.py
daniaki: Testing without the filter first sounds like a good idea, then if we need to slim the results down we can add a filter of 0.1 like GTEx.

KatalinaBobowik: Sounds great, I'll test that one first :)

KatalinaBobowik:
@katiedelange @daniaki @danielreti I removed the filtering step on the ld calculation and reverted the chromosome field in the association output back to the 'chr' prefix. Tests are also all running sucessfully (including the wrapper script 🎉). Let me know what you think.

katiedelange (approved):
Looks great, thanks Kat! 

Only one general comment for all of us to consider more broadly: several of these changes are to adjust whether CSV or TSV is used as input/output. Going forward it would probably be good if we could all agree on a consistent format that we will use (as much as possible) to pass this sort of data back and forth.

daniaki:
Thanks Kat! In reference to @katiedelange's point above, my preference goes to TSV (gzipped to save space for large files) as outputs to avoid issues when commas appear in column values.

KatalinaBobowik:
great, thanks @daniaki. Maybe we can set that as the default then? I'll make sure to use TSV going forward

PR #99, opened 2022-07-18
Change access level to 'standard' if bucket name is 'main'

KatalinaBobowik:
Previously, I was testing data using the `test` bucket, however found a mistake when running in `main`. 

Data access can be either `standard`, `full`, or `test`. When previously running the script, I used the gs://cpg-<dataset>-{main,test} syntax, using `test` as the access level. However, this does not work for main, as access levels are `standard` or `full`.

### DIFF
--- a/scripts/eqtl_hail_batch/launch_generate_eqtl_spearman.py
+++ b/scripts/eqtl_hail_batch/launch_generate_eqtl_spearman.py
@@ -129,6 +129,8 @@ def submit_eqtl_jobs(
                 analysis_runner_output_path = output_prefix[5:].partition('/')[-1]
                 # get access level from bucket, rather than manual input
                 access_level = bucket_name.split('-')[-1]
+                if access_level == 'main':
+                    access_level = 'standard'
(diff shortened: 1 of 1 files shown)


katiedelange (approved):

scripts/eqtl_hail_batch/launch_generate_eqtl_spearman.py
```diff
@@ -129,6 +129,8 @@ def file_exists(files):
                 analysis_runner_output_path = output_prefix[5:].partition('/')[-1]
                 # get access level from bucket, rather than manual input
                 access_level = bucket_name.split('-')[-1]
+                if access_level == 'main':
+                    access_level = 'standard'
```
lgruen, line 133: It's cleaner if you don't derive this from the bucket name, but instead use the config value, so you don't rely on a particular bucket naming scheme:

```python
from cpg_utils.hail_batch import get_config
...
access_level = get_config()['workflow']['access_level']
```

PR #103, opened 2022-07-26
Remove hom refs

KatalinaBobowik:
Optimise efficiency of association analysis (minimise rekeying, repartitioning, and checkpointing). I've also removed an error where a locus has no variation in the number of alternative alleles.

### DIFF
--- a/scripts/eqtl_hail_batch/generate_eqtl_spearman.py
+++ b/scripts/eqtl_hail_batch/generate_eqtl_spearman.py
@@ -458,20 +458,23 @@ def run_spearman_correlation_scatter(
         """get Spearman rank correlation"""
         gene_symbol = df.gene_symbol
         gene_id = df.gene_id
+        alleles = df.alleles
         snp = df.snpid
+        print(f'Printing SNP: {snp}')
         gt = genotype_df[genotype_df.snpid == snp][['sampleid', 'n_alt_alleles']]
         res_val = residuals_df[['sampleid', gene_symbol]]
         test_df = res_val.merge(gt, on='sampleid', how='right')
         test_df.columns = ['sampleid', 'residual', 'SNP']
         # set spearmanr calculation to perform the calculation ignoring nan values
         coef, p = spearmanr(test_df['SNP'], test_df['residual'], nan_policy='omit')
-        return (gene_symbol, gene_id, snp, coef, p)
+        return (gene_symbol, gene_id, alleles, snp, coef, p)
 
     # run spearman correlation function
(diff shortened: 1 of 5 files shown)


scripts/eqtl_hail_batch/launch_generate_eqtl_spearman.py
```diff
@@ -114,40 +117,39 @@ def file_exists(files):
...
-                    # commit, sha and cwd can be inferred automatically
-                    script=[
-                        'generate_eqtl_spearman.py',
-                        *('--expression', expression),
-                        *('--covariates', covariates),
-                        *('--geneloc', geneloc),
-                        *('--keys', keys),
-                        *('--output-prefix', output_prefix),
-                    ],
+                job.command(
```
lgruen, line 142: This looks fine to me, but @illusional suggested that it might be easier to call this directly.

Is this performing much work? If not, could you just call a Python function from this module inline here to construct a job for a larger batch, instead of the indirection through either another batch or calling `subprocess`?

Certainly not a big issue though, so feel free to ignore / revisit in follow-up PR.

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -488,21 +488,15 @@ def spearman_correlation(df):
...
-    t = t.key_by('locus')
-    t = t.annotate(
-        a1=mt.rows()[t.locus].alleles[0],
-        a2=hl.if_else(
-            hl.len(mt.rows()[t.locus].alleles) == 2, mt.rows()[t.locus].alleles[1], 'NA'
-        ),
-    )
     # add in vep annotation
-    t = t.annotate(functional_annotation=mt.rows()[t.locus].vep_functional_anno)
+    t = t.key_by('locus', 'alleles')
```
lgruen, line 492: This will cause a shuffle, but I guess there's no way around this, as you want to run the `spearman_correlation` function outside of Hail, which presumably can't be expressed easily as a Hail expression.

lgruen (approved):

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
KatalinaBobowik: Correct, this is one of the few instances where I use `key_by()`, but it's because the output of the association analysis is a df

scripts/eqtl_hail_batch/launch_generate_eqtl_spearman.py
KatalinaBobowik: If this is working ok I might leave as is, then ask @illusional about it for a follow-up PR

PR #105, opened 2022-07-26
Fix output path

KatalinaBobowik:
I'm running into an issue where I would like to add a version number to my output files (in case a file gets corrupted/changes, etc) however I'm currently using `dataset_path` as the prefix of my output file, since `output_path` changes per cell type and I only need to generate this file once. I've added in a version to the output table, which is taken from the input-prefix.

@lgruen this is a crude way of fixing this. Please let me know if you think there's a better way

### DIFF
--- a/scripts/eqtl_hail_batch/launch_generate_eqtl_spearman.py
+++ b/scripts/eqtl_hail_batch/launch_generate_eqtl_spearman.py
@@ -44,13 +44,13 @@ from google.cloud import storage
     '--input-path',
     required=True,
     help=(
-        'A path prefix of where input files are located. Version number must be included on the end. eg: gs://MyBucket/folder/v0. '
+        'A path prefix of where input files are located. eg: gs://MyBucket/folder. '
     ),
 )
 @click.option(
     '--output-dir',
     required=True,
-    help='A path of where to output files. eg: gs://MyBucket/output-folder/',
+    help='A path of where to output files. Version number must be included on the end. eg: gs://MyBucket/output-folder/v0',
 )
 @click.option('--dry-run', is_flag=True, help='Just check if files exist')
 def submit_eqtl_jobs(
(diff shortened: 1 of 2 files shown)


scripts/eqtl_hail_batch/launch_generate_eqtl_spearman.py
```diff
@@ -136,8 +136,10 @@ def submit_eqtl_jobs(
                 )
 
                 keys = os.path.join(input_path, 'OneK1K_CPG_IDs.tsv')
+                version = output_dir.split('/')[-1]
```
lgruen: An alternative approach would be to not launch this script with the output path containing the cell type, but instead rely on the script to append the cell type to the path itself when you write outputs. Does that make sense?

scripts/eqtl_hail_batch/launch_generate_eqtl_spearman.py
```diff
@@ -136,8 +136,10 @@ def submit_eqtl_jobs(
...
+                # put version number at end of output path
                 output_prefix = os.path.join(
-                    output_dir, f'{cell_type}', f'chr{chromosome}'
+                    output_dir.strip(version), f'{cell_type}', f'chr{chromosome}', f'{version}'
```
lgruen: `strip` is probably not what you want here, but `removesuffix`. (See https://stackabuse.com/python-remove-the-prefix-and-suffix-from-a-string/.)

lgruen (changes requested):

scripts/eqtl_hail_batch/launch_generate_eqtl_spearman.py
KatalinaBobowik: Yes, that's a good idea. And would look a lot cleaner so I could just use `output_path` instead

scripts/eqtl_hail_batch/launch_generate_eqtl_spearman.py
KatalinaBobowik: neat, I didn't know about `removesuffix`

lgruen (approved):

PR #108, opened 2022-08-03
add reference boundaries to filter interval

MattWellie:
When building a filter interval, always stay within the reference genome boundaries

https://centrepopgen.slack.com/archives/C018KFBCR1C/p1659570003763679?thread_ts=1658208231.938539&cid=C018KFBCR1C


KatalinaBobowik (approved):
Thanks for the suggestion, Matt! Looks great

PR #110, opened 2022-08-17
Rare variant association

annacuomo:


scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,168 @@
+import os
+import sys
+import click
+import logging
+import hail as hl
+import pandas as pd
+from cpg_utils.hail_batch import output_path
+
```
MattWellie, line 9:
```suggestion
from cloudpathlib import AnyPath

```

scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,168 @@
...
+    sample_key_filename = (
+        'gs://cpg-tob-wgs-test/scrna-seq/grch38_association_files/OneK1K_CPG_IDs.tsv'
+    )
+    sample_key_df = pd.read_csv(sample_key_filename, sep='\t', index_col=0)
+
+    cpg_id = sample_key_df[sample_key_df['OneK1K_ID'] == onek1k_id]['InternalID']
+    logging.info('CPG ID: {}'.format(cpg_id))  # e.g., 'CPG9951'
+
+    # define output filename and check if it already exists
+    output_filename = output_path(f'{cpg_id}_{gene_name}.csv')
```
MattWellie:
```suggestion
    output_file = AnyPath(output_path(f'{cpg_id}_{gene_name}.csv'))
```

scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,168 @@
...
+    logging.info('Output file: {}'.format(output_filename))
+
+    # skip if file already exists
+    if os.path.exists(output_filename):
+        logging.info('File already exists, exiting')
+        sys.exit()
```
MattWellie:
```suggestion
    if output_file.exists():
        raise Exception(f'File {output_filename} already exists, exiting')
```

scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,168 @@
...
+        'maf_onek1k': maf_sample_list,
+        'maf_gnomad_popmax': maf_popmax,
+        'regulatory_consequences': regulatory_consequences,
+        'ensembl_gene_id': ensembl_gene_id,
+        'chrom': chrom,
+        'gene_start': int(gene_start),
+        'gene_end': int(gene_end),
+    }
+    df = pd.DataFrame(results_data)
+    df.to_csv(output_filename)
```
MattWellie:
```suggestion
    df.to_csv(output_file.open())
```

scripts/rv_expression_association/get_rv_outliers.py
MattWellie: AnyPath means that you can check for/open any type of path. `os.path` won't work for cloud paths

MattWellie (changes requested):
A few changes to do with the file interaction. After that I think it should work fine!

scripts/rv_expression_association/get_rv_outliers.py
annacuomo: does the exception also "exit"?

scripts/rv_expression_association/get_rv_outliers.py
annacuomo: ok, thank you!

scripts/rv_expression_association/get_rv_outliers.py
annacuomo: also note to self to remove `import sys` if that line is not needed anymore

MattWellie: Yup! Raising an Exception will cause everything to stop, unless you've deliberately captured everything in a try-except
https://www.w3schools.com/python/python_try_except.asp

annacuomo: ok perfect!

scripts/rv_expression_association/README.md
```diff
@@ -0,0 +1,39 @@
+## Outlier analysis
+
+### Generate positive association examples
+The first step for this analysis is to construct a set of "true associations" by identifying expression outliers (_e.g.,_ individuals that have extreme - outlier - expression for a specific gene) and then identifying genetic variants within the vicinity of that gene for which the outlier individual(s) have alternative alleles (either 0/1 or 1/1).
+
+[This script](../rv_expression_association/get_rv_outliers.py) takes a gene and an individual IDs as inputs and returns the following file:
```
MattWellie:
```suggestion
[This script](get_rv_outliers.py) takes a gene and an individual IDs as inputs and returns the following file:
```
Nit: I think this extra path is redundant...

scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,166 @@
...
+
+    create the following table:
+    donor ID | gene ID | variant ID | position | CADD | MAF (OneK1K) | MAF (gnomad) | ..
+    """
+
+    # file matching OneK1K IDs to CPG (internal) and TOB (external) IDs
+    sample_key_filename = (
+        'gs://cpg-tob-wgs-test/scrna-seq/grch38_association_files/OneK1K_CPG_IDs.tsv'
+    )
+    sample_key_df = pd.read_csv(sample_key_filename, sep='\t', index_col=0)
```
MattWellie: Can you test-run this to make sure the syntax works? I don't think pandas will be able to automatically resolve gs:// paths, but I could be wrong..

scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,166 @@
...
+
+    # check this file in the future (GENCODE??)
+    # get gene body position (start and end) and build interval
+    # include variants up to 10kb up- and downstream
+    gene_file = (
+        'gs://cpg-tob-wgs-test/scrna-seq/grch38_association_files/gene_location_files/GRCh38_geneloc_chr'
+        + chrom
+        + '.tsv'
+    )
+    gene_df = pd.read_csv(gene_file, sep='\t', index_col=0)
```
MattWellie: same here RE: pandas opening cloudpaths

scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,166 @@
...
+    interval_start = float(gene_df[gene_df['gene_name'] == gene_name]['start']) - 10000
+    interval_end = float(gene_df[gene_df['gene_name'] == gene_name]['end']) + 10000
+
+    # get gene-specific genomic interval
+    gene_interval = (
+        'chr' + chrom + ':' + str(interval_start) + '-' + str(interval_end)
+    )  # 'chr22:23219960-23348287'
```
MattWellie:
```suggestion
    gene_interval = f'chr{chrom}:{interval_start}-{interval_end}'
    # 'chr22:23219960-23348287'
```
Woooo f-strings! They are sexy, _and_ will automatically do type conversions so you don't need to wrap objects in `str()`

scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,166 @@
...
+    donor_mt = hl.filter_intervals(
+        donor_mt, [hl.parse_locus_interval(gene_interval, reference_genome='GRCh38')]
+    )
+    logging.info('Number of variants within interval: {}'.format(donor_mt.count()[0]))
+
+    # remove variants for which this individual is 0/0
+    donor_mt = hl.variant_qc(donor_mt)
+    donor_mt = donor_mt.filter_rows(donor_mt.variant_qc.n_non_ref > 0)
+    logging.info(
+        'Number of non-ref variants for this indvidual: {}'.format(donor_mt.count()[0])
```
MattWellie:
```suggestion
        f'Number of non-ref variants for this indvidual: {donor_mt.count()[0]}'
```
Wooooo f-strings!

MattWellie (changes requested):
Sorry, a couple more things. The changes from `format` to f-strings are not functional ones, but pandas file reading is something I'm not sure about. Those file paths may need to be AnyPath('filenam').open() for pandas to access the content

scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,192 @@
...
+    - CADD score
+    - MAF within the OneK1K sample
+    - MAF in gnomad
+
+    create the following table:
+    donor ID | gene ID | variant ID | position | CADD | MAF (OneK1K) | MAF (gnomad) | ..
+    """
+
+    # file matching OneK1K IDs to CPG (internal) and TOB (external) IDs
+    sample_key_filename = (
```
illusional: Can this be an input to the script?

scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,192 @@
...
+        )
+    )
+
+    # retain info about het variants
+    is_het_list = donor_mt.variant_qc.n_het.collect()
+
+    # annotate variants with CADD scores, gnomad etc
+    logging.info('Annotate variants with CADD scores and gnomad AF')
+    ref_ht = hl.read_table(
+        'gs://cpg-reference/seqr/v0-1/combined_reference_data_grch38-2.0.4.ht'
```
illusional: This makes this this path more portable across environments:

```python
from cpg_utils.hail_batch import reference_path

reference_path('seqr/v0-1/combined_reference_data_grch38-2.0.4.ht')
```

illusional (changes requested):
Few requests:

- Make any specific paths an input to the script
- We tend to use f-strings (`f'my {input_string}'`) over formatting `'my {}'.format(input_string)`
- Use the `reference_path` function to select references to make this script more portable.

MattWellie (dismissed):

scripts/rv_expression_association/debug_maf_distributions.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
+
+import hail as hl
+# from bokeh.io import show
+from bokeh.io.export import get_screenshot_as_png
+from cpg_utils.hail_batch import output_path, init_batch 
+
+init_batch()
+mt = hl.read_matrix_table('gs://cpg-tob-wgs-test/tob_wgs_vep/v1/vep105_GRCh38.mt')
```
lgruen: As discussed on Slack, this should point to `v7.mt` and use `dataset_path`, as the VEP annotated version is broken w.r.t. densification.

scripts/rv_expression_association/debug_maf_distributions.py
```diff
@@ -0,0 +1,30 @@
...
+mt = hl.experimental.densify(mt)
```
lgruen: If you wanted, you could at least filter to the chromosome that your target interval is on, to speed densification up.

scripts/rv_expression_association/debug_maf_distributions.py
```diff
@@ -0,0 +1,30 @@
...
+mt = hl.variant_qc(mt)
+mt = hl.filter_intervals(
+    mt, [hl.parse_locus_interval('chr22:23704425-23802743', reference_genome='GRCh38')]
+)
+mt = mt.filter_rows(hl.len(hl.or_else(mt.filters, hl.empty_set(hl.tstr))) == 0)
```
lgruen: It would be good to add a `checkpoint` after this line, as otherwise all operations above will be reexecuted for line 23 and after.

scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,183 @@
...
+@click.option('--gene-file', required=True)
+@click.option(
+    '--window-size',
+    required=True,
+    help='the size of the flanking region, to be added twice, on either side of the gene body',
+)
+def main(
+    onek1k_id: str,
+    gene_name: str,
+    sample_mapping_file: str,  # 'gs://cpg-tob-wgs-test/scrna-seq/grch38_association_files/OneK1K_CPG_IDs.tsv'
```
lgruen: Can you map this through `dataset_path` instead, so you don't need to adjust the parameters when running in different access levels?

scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,183 @@
...
+    - CADD score
+    - MAF within the OneK1K sample
+    - MAF in gnomad
+
+    create the following table:
+    donor ID | gene ID | variant ID | position | CADD | MAF (OneK1K) | MAF (gnomad) | ..
+    """
+
+    # file matching OneK1K IDs to CPG (internal) and TOB (external) IDs
+    sample_key_df = pd.read_csv(AnyPath(sample_mapping_file), sep='\t', index_col=0)
```
lgruen: @illusional We don't have both of these external IDs in Metamist, I assume?

scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,183 @@
...
+    output_filename = AnyPath(output_path(f'{cpg_id}_{gene_name}_{window_size}.csv'))
+    logging.info(f'Output file: {output_filename}')
+
+    # skip if file already exists
+    if output_filename.exists():
+        raise Exception(f'File {output_filename} already exists, exiting')
+
+    init_batch()
+    # get VEP-annotated WGS object (hail matrix table)
+    mt = hl.read_matrix_table('gs://cpg-tob-wgs-test/tob_wgs_vep/v1/vep105_GRCh38.mt')
```
lgruen: Ditto.

scripts/rv_expression_association/plot/plot2.py
```diff
@@ -0,0 +1,45 @@
...
+    init_batch()
+
+    mt = hl.read_matrix_table(VEP_MT)
+    mt = hl.filter_intervals(
+        mt,
+        [hl.parse_locus_interval('chr22:23000000-24000000', reference_genome='GRCh38')],
+    )
+    mt = hl.experimental.densify(mt)
+
+    mt = hl.variant_qc(mt)
```
lgruen: Ditto about checkpoint.

lgruen (changes requested):

scripts/rv_expression_association/get_rv_outliers.py
illusional: I don't think we have OneK1K IDs, or at least I don't remember seeing them. W would want to address https://github.com/populationgenomics/sample-metadata/issues/139 before we ingested them. But seems relatively important.

scripts/rv_expression_association/debug_maf_distributions.py
annacuomo: i think this script is not needed anymore so will delete, but adopting the corrections for get_rv_outliers.py :)

scripts/rv_expression_association/debug_maf_distributions.py
annacuomo: ok! Just trying to figure out the syntax now :)

scripts/rv_expression_association/debug_maf_distributions.py
annacuomo: i've tried adding a checkpoint (in the `get_rv_outlier.py` script) at lines 85-86 but not super sure of the syntax!

scripts/rv_expression_association/get_rv_outliers.py
annacuomo: you mean instead of passing it as an argument? Or in the script, instead of (or in addition of) AnyPath??

annacuomo: so am i passing e.g., `scrna-seq/grch38_association_files/OneK1K_CPG_IDs.tsv` instead of `gs://cpg-tob-wgs-test/scrna-seq/grch38_association_files/OneK1K_CPG_IDs.tsv` as an argument, and then change line 52 (of get_rv_outliers.py) to `sample_key_df = pd.read_csv(AnyPath(dataset_path(sample_mapping_file)), sep='\t', index_col=0)` ??? @lgruen

scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,197 @@
...
+    # mt = hl.read_matrix_table('gs://cpg-tob-wgs-test/tob_wgs_vep/v1/vep105_GRCh38.mt')
+    mt = hl.read_matrix_table(MT)
+    logging.info(f'Number of total variants: {mt.count()[0]}')
+
+    # get relevant chromosome
+    gene_df = pd.read_csv(AnyPath(dataset_path(gene_file)), sep='\t', index_col=0)
+    chrom = gene_df[gene_df['gene_name'] == gene_name]['chr']
+
+    # filter to relevant chromosome to speed densification up
+    mt = mt.filter_rows(mt.locus.chrom == ('chr'+chrom))  # figure out actual syntax
```
lgruen: It's `locus.contig`: https://hail.is/docs/0.2/genetics/hail.genetics.Locus.html#hail.genetics.Locus

scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,197 @@
...
+    logging.info(f'Number of variants on chromosome {chrom}: {mt.count()[0]}')
+
+    # densify
+    mt = hl.experimental.densify(mt)
+
+    # filter out low QC variants
+    mt = mt.filter_rows(hl.len(hl.or_else(mt.filters, hl.empty_set(hl.tstr))) == 0)
+    mt_path = output_path('densified_chrom_and_qc_filtered.mt', 'tmp')
+    mt = mt.checkpoint(mt_path)
```
lgruen: Consider adding `overwrite=True` here if you want to run this multiple times.

lgruen (changes requested):

scripts/rv_expression_association/get_rv_outliers.py
lgruen: You could either pass `scrna-seq/grch38_association_files/OneK1K_CPG_IDs.tsv` as a parameter and use `dataset_path(sample_mapping_file)` to get the full path based on the access level, or if it's unlikely to ever change, you could just define it as a constant, too.

scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,197 @@
...
+@click.option(
+    '--window-size',
+    required=True,
+    help='the size of the flanking region, to be added twice, on either side of the gene body',
+)
+def main(
+    onek1k_id: str,
+    gene_name: str,
+    sample_mapping_file: str,  # 'scrna-seq/grch38_association_files/OneK1K_CPG_IDs.tsv'
+    gene_file: str,  # e.g., 'scrna-seq/grch38_association_files/gene_location_files/GRCh38_geneloc_chr1.tsv'
```
lgruen, line 31: How about using https://hail.is/docs/0.2/experimental/index.html#hail.experimental.import_gtf instead?

scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,197 @@
...
+    donor_mt = mt.filter_cols(mt.s == cpg_id)
+
+    # get gene body position (start and end) and build interval
+    # include variants up to {window size} up- and downstream
+    interval_start = int(gene_df[gene_df['gene_name'] == gene_name]['start']) - int(
+        window_size
+    )
+    interval_end = int(gene_df[gene_df['gene_name'] == gene_name]['end']) + int(
+        window_size
+    )
```
lgruen, line 101: You'll have to clip these against the chromosome boundaries: https://github.com/populationgenomics/tob-wgs/pull/108 (@MattWellie)

scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,197 @@
...
+    donor_mt = hl.variant_qc(donor_mt)
+    donor_mt = donor_mt.filter_rows(donor_mt.variant_qc.n_non_ref > 0)
+    logging.info(
+        f'Number of non-ref variants for this indvidual: {donor_mt.count()[0]}'
+    )
+
+    # focus on SNVs for now
+    donor_mt = donor_mt.filter_rows(donor_mt.vep.variant_class == 'SNV')
+    # filter for biallelic only
+    donor_mt = donor_mt.filter_rows(hl.len(donor_mt.alleles) == 2)  # not needed?
```
lgruen: The input file already has multi-allelics split.

scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,197 @@
...
+    logging.info(f'Output file: {output_filename}')
+
+    # skip if file already exists
+    if output_filename.exists():
+        raise Exception(f'File {output_filename} already exists, exiting')
+
+    init_batch()
+    # get VEP-annotated WGS object (hail matrix table)
+    # mt = hl.read_matrix_table('gs://cpg-tob-wgs-test/tob_wgs_vep/v1/vep105_GRCh38.mt')
+    mt = hl.read_matrix_table(MT)
```
lgruen, line 71: At least the currently set `MT` doesn't have VEP annotations.

scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,197 @@
...
+
+    # remove variants for which this individual is 0/0
+    donor_mt = hl.variant_qc(donor_mt)
+    donor_mt = donor_mt.filter_rows(donor_mt.variant_qc.n_non_ref > 0)
+    logging.info(
+        f'Number of non-ref variants for this indvidual: {donor_mt.count()[0]}'
+    )
+
+    # focus on SNVs for now
+    donor_mt = donor_mt.filter_rows(donor_mt.vep.variant_class == 'SNV')
```
lgruen: You could also use Hail's `is_snp` to filter, potentially even before annotating with VEP.

scripts/rv_expression_association/get_rv_outliers.py
```diff
@@ -0,0 +1,197 @@
...
+        'cadd': cadd_list,
+        'maf_onek1k': maf_sample_list,
+        'maf_gnomad_popmax': maf_popmax,
+        # 'regulatory_consequences': regulatory_consequences,
+        'ensembl_gene_id': ensembl_gene_id,
+        'chrom': chrom,
+        'gene_start': int(gene_start),
+        'gene_end': int(gene_end),
+    }
+    df = pd.DataFrame(results_data)
```
lgruen, line 200: Why not write an annotated Hail table instead? Collecting all data on a single node doesn't scale well -- also CSV is not a great format for a number of reasons (lack of a standard, highly inefficient serialization / deserialization as it's not binary-based, loss of numeric precision, etc.).

lgruen (changes requested):

scripts/rv_expression_association/get_rv_outliers.py
MattWellie: reference to this?
https://github.com/populationgenomics/tob-wgs/blob/main/scripts/eqtl_hail_batch/generate_eqtl_spearman.py#L351-L360

scripts/rv_expression_association/get_rv_outliers.py
lgruen: https://hail.is/docs/0.2/experimental/index.html#hail.experimental.get_gene_intervals might be even easier.

scripts/rv_expression_association/get_rv_outliers.py
annacuomo: Right ok! I was never going to run this at scale, just generating this for a few selected examples, trying to annotate a few metrics to check the distributions of etc, and I thought a CSV would be pretty general, can be opened with R as well for interesting plots etc. Once I am set on the specific metric values to filter on etc I was going to convert the (tightly subsetted) hail object to plink format instead (until i change my method to take in hail objects directly). But if you feel strongly about this i can use an annotated hail table instead, and just get better at hail plotting :)

scripts/rv_expression_association/get_rv_outliers.py
annacuomo: ok! so biallelic filter is not needed (correct?) and swap this line for `is_snp`?

scripts/rv_expression_association/get_rv_outliers.py
annacuomo: as in all variants included here are biallelic already?

scripts/rv_expression_association/get_rv_outliers.py
annacuomo: i will want to run this once we have them again though I think, can that annotation be run again at all?

scripts/rv_expression_association/get_rv_outliers.py
annacuomo: ok, i've done the former, assuming that i still also need AnyPath for pandas to find it?

scripts/rv_expression_association/get_rv_outliers.py
annacuomo: perfect thanks!

scripts/rv_expression_association/get_rv_outliers.py
annacuomo: sorry, could you provide more detail? I thought this worked fine previously?

scripts/rv_expression_association/get_rv_outliers.py
annacuomo: that would be just to find the start / end of the gene, I'd still need to define the interval?

annacuomo: ah you're saying as an alternative to the gene_file! That would work actually, I'd also not need to specify the right chromosome every time! Do we have access to a Gencode GTF?

lgruen: Yes, we do, however there's some related discussion on how best to read it: https://centrepopgen.slack.com/archives/C018KFBCR1C/p1661409100527349?thread_ts=1658208231.938539&cid=C018KFBCR1C

scripts/rv_expression_association/get_rv_outliers.py
lgruen: If by extending the interval around the gene you go out of chromosome / contig boundaries, the filtering step below will fail (e.g. imagine you're asking Hail to filter from (negative!) position -100 to 400 or so).

scripts/rv_expression_association/get_rv_outliers.py
lgruen: Yes, for `v7.mt`, everything has been split to biallelics already.

scripts/rv_expression_association/get_rv_outliers.py
lgruen: Yes, certainly, that's probably something you should coordinate with @KatalinaBobowik. There's also a discussion about the GENCODE / VEP version here: https://centrepopgen.slack.com/archives/C018KFBCR1C/p1661468741596349?thread_ts=1658208231.938539&cid=C018KFBCR1C

scripts/rv_expression_association/get_rv_outliers.py
lgruen: Yup!

scripts/rv_expression_association/get_rv_outliers.py
lgruen: Sounds fine -- in general, please treat all review comments as suggestions! Certainly not trying to push you to tools that aren't fit for the use case.

FWIW there's also some Hail / R interoperability libraries. I haven't looked at them in detail, but I think you should be able to work with a Hail table directly in R this way, without converting to PLINK first.

scripts/rv_expression_association/get_rv_outliers.py
annacuomo: ah yes ok that makes sense! Will have a look at Matt's code :)

lgruen (changes requested):
Please let me know when this is ready for another look.

PR #116, opened 2022-08-24
Add ensembl ids

KatalinaBobowik:
Add ENSEMBL IDs to expression output data

### DIFF
--- a/scripts/eqtl_hail_batch/generate_eqtl_spearman.py
+++ b/scripts/eqtl_hail_batch/generate_eqtl_spearman.py
@@ -27,7 +27,7 @@ assert MULTIPY_IMAGE
 
 TOB_WGS = dataset_path('mt/v7.mt/')
 FREQ_TABLE = dataset_path('joint-calling/v7/variant_qc/frequencies.ht/', 'analysis')
-VEP_ANNOTATION = dataset_path('tob_wgs_vep/v1/vep105_GRCh38.mt/')
+VEP_ANNOTATION = dataset_path('tob_wgs_vep/104/vep104.3_GRCh38.ht/')
 GENCODE_GTF = 'gs://cpg-reference/gencode/gencode.v38.annotation.gtf.bgz'
 
 
@@ -177,7 +177,7 @@ def generate_log_cpm_output(expression_df, output_prefix, celltype):
     # add in cell type info
     data_summary['cell_type_id'] = celltype
     # add in ENSEMBL IDs
-    gtf = hl.experimental.import_gtf(GENCODE_GTF, force=True, reference_genome='GRCh38', skip_invalid_contigs=True)
+    gtf = hl.experimental.import_gtf(GENCODE_GTF, reference_genome='GRCh38', skip_invalid_contigs=True)
     data_summary['ensembl_ids'] = data_summary.merge(gtf.drop_duplicates('gene_name'), how='left', on='gene_name').gene_id
 
     # Save file
--- a/scripts/eqtl_hail_batch/generate_eqtl_spearman.py
+++ b/scripts/eqtl_hail_batch/generate_eqtl_spearman.py
@@ -4,6 +4,7 @@
 
 
 import hail as hl
+import re
 import hailtop.batch as hb
 import pandas as pd
 import numpy as np
@@ -28,6 +29,7 @@ assert MULTIPY_IMAGE
 TOB_WGS = dataset_path('mt/v7.mt/')
 FREQ_TABLE = dataset_path('joint-calling/v7/variant_qc/frequencies.ht/', 'analysis')
 VEP_ANNOTATION = dataset_path('tob_wgs_vep/v1/vep105_GRCh38.mt/')
+GENCODE_GTF = 'gs://cpg-reference/gencode/gencode.v38.annotation.gtf.gz'
 
 
 def filter_lowly_expressed_genes(expression_df):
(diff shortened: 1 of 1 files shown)


scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -175,6 +177,13 @@ def create_struct(gene):
     data_summary = data_summary.rename({'index': 'gene_symbol'}, axis='columns')
     # add in cell type info
     data_summary['cell_type_id'] = celltype
+    # add in ENSEMBL IDs
+    gtf = pd.read_csv(GENCODE_GTF, sep='\t', header=None, index_col=False, skiprows=5)
```
lgruen: Would it be easier to use https://hail.is/docs/0.2/experimental/index.html#hail.experimental.import_gtf here instead?

KatalinaBobowik: Doesn't that get imported as a hail table? I only need a pandas df here

lgruen: Yes, but you could easily convert the Hail Table to a Pandas DataFrame. You then wouldn't need any of that parsing code below.

KatalinaBobowik: Would you happen to know how to import the GTF? Using the same GTF file I specified at the top of the code and loading in using
```
hl.experimental.import_gtf(GENCODE_GTF, reference_genome='GRCh38', skip_invalid_contigs=True, force_bgz=True)
```
I get 
```
Error summary: ZipException: File does not conform to block gzip format.
```

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -28,6 +29,7 @@
 TOB_WGS = dataset_path('mt/v7.mt/')
 FREQ_TABLE = dataset_path('joint-calling/v7/variant_qc/frequencies.ht/', 'analysis')
 VEP_ANNOTATION = dataset_path('tob_wgs_vep/v1/vep105_GRCh38.mt/')
+GENCODE_GTF = 'gs://cpg-reference/gencode/gencode.v38.annotation.gtf.gz'
```
lgruen: Unrelated to this PR, but noting that we should likely rerun this with VEP 104?

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
lgruen: I think this is still pending on the discussion here: https://centrepopgen.slack.com/archives/C018KFBCR1C/p1661410430973899?thread_ts=1658208231.938539&cid=C018KFBCR1C

lgruen (changes requested):

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
KatalinaBobowik: I'm happy to rerun VEP using v104 if that's what's decided upon

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
KatalinaBobowik: This works using `force = True`, and I've added this in now.

KatalinaBobowik:
Just to confirm, have we decide on Gencode v 38? If so, I just need the correct link, as `gs://cpg-reference/gencode/gencode.v38.annotation.gtf.gz` currently doesn't exist

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -28,6 +28,7 @@
 TOB_WGS = dataset_path('mt/v7.mt/')
 FREQ_TABLE = dataset_path('joint-calling/v7/variant_qc/frequencies.ht/', 'analysis')
 VEP_ANNOTATION = dataset_path('tob_wgs_vep/v1/vep105_GRCh38.mt/')
```
lgruen: This should point to `vep104`, right?

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -172,9 +173,12 @@ def create_struct(gene):
     data_summary = pd.DataFrame(
         log_cpm.apply(create_struct, axis=0), columns=['data']
     ).reset_index()
-    data_summary = data_summary.rename({'index': 'gene_symbol'}, axis='columns')
+    data_summary = data_summary.rename({'index': 'gene_name'}, axis='columns')
     # add in cell type info
     data_summary['cell_type_id'] = celltype
+    # add in ENSEMBL IDs
+    gtf = hl.experimental.import_gtf(GENCODE_GTF, force=True, reference_genome='GRCh38', skip_invalid_contigs=True)
```
lgruen: Now that we have `.bgz`, you shouldn't need the `force=True` anymore.

lgruen (approved):

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
KatalinaBobowik: Ok, good to know. I've removed it now

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
KatalinaBobowik: Yes, thanks for that. I've fixed the path

PR #118, opened 2022-09-01
Fix oom error

KatalinaBobowik:
Fixing OOM error due to serializing data frames. This now runs successfully in the `test` bucket.

### DIFF
--- a/scripts/eqtl_hail_batch/conditional_analysis.py
+++ b/scripts/eqtl_hail_batch/conditional_analysis.py
@@ -240,8 +240,8 @@ def run_computation_in_scatter(
     )
     # save esnp1 for front-end use on which SNPs have been conditioned on
     esnp1_path = AnyPath(output_prefix) / f'conditioned_esnps_{iteration}.tsv'
-    with esnp1_path.open('w') as fp:
-        esnp1.to_csv(fp, index=False)
+    # with esnp1_path.open('w') as fp:
+    #     esnp1.to_csv(fp, index=False)
 
     # Remaning eSNPs to test
     esnps_to_test = (
@@ -335,7 +335,7 @@ def run_computation_in_scatter(
         lambda x: ':'.join(x[['chrom', 'bp', 'a1', 'a2']]), axis=1
     )
 
-    # TODO Kat to fix
+    # save each sig snps file as a parquet
     output_path = os.path.join(output_prefix, f'sig-snps-{idx}.parquet')
     adjusted_spearman_df.to_parquet(output_path)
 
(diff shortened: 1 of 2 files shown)


scripts/eqtl_hail_batch/conditional_analysis.py
```diff
@@ -77,17 +77,20 @@ def get_genotype_df(residual_df, gene_snp_test_df):
     # only keep samples that are contained within the residuals df
     # this is important, since not all indivuduals have expression/residual
     # data (this varies by cell type)
-    samples_to_keep = set(residual_df.sampleid)
-    set_to_keep = hl.literal(samples_to_keep)
-    mt = mt.filter_cols(set_to_keep.contains(mt['onek1k_id']))
+    samples_to_keep = hl.literal(list(residual_df.sampleid))
```
lgruen: `contains` is likely faster on a `set` compared to a list. Not sure if you need to first convert to a list (`hl.set(list(residual_df.sampleid))`) though.

```suggestion
    samples_to_keep = hl.set(residual_df.sampleid)
```

scripts/eqtl_hail_batch/conditional_analysis.py
```diff
@@ -316,43 +330,37 @@ def spearman_correlation(df):
...
+            x[['chrom', 'bp', 'a1', 'a2', 'gene_symbol', 'cell_type_id', 'round']]
+        ),
+        axis=1,
+    )
+    adjusted_spearman_df['variant_id'] = adjusted_spearman_df.apply(
+        lambda x: ':'.join(x[['chrom', 'bp', 'a2']]), axis=1
+    )
+    adjusted_spearman_df['snp_id'] = adjusted_spearman_df.apply(
+        lambda x: ':'.join(x[['chrom', 'bp', 'a1', 'a2']]), axis=1
+    )
```
lgruen: Didn't we want to remove those IDs?

lgruen (changes requested):
Did you already test this with the `v2` test inputs?

scripts/eqtl_hail_batch/generate_eqtl_spearman.py
```diff
@@ -656,7 +661,8 @@ def main(
     copy_common_env(filter_mt_job)
     filtered_mt_path = filter_mt_job.call(prepare_genotype_info, keys_path=keys)
 
-    for idx in range(n_genes_in_scatter):
+    # for idx in range(n_genes_in_scatter):
```
illusional: Should this be reverted?

KatalinaBobowik: yes, good catch, thanks! I'll update that

scripts/eqtl_hail_batch/conditional_analysis.py
KatalinaBobowik: Sorry yes, I forgot about those. That's removed now

scripts/eqtl_hail_batch/conditional_analysis.py
KatalinaBobowik: `hl.set(list(residual_df.sampleid))` works 👌

lgruen (approved):

PR #121, opened 2022-09-05
Batchification of EQTL analysis

illusional:
Due to the desire to pull out certain stages, it's really reached the point for this to be one bigger hail batch workflow. Going through this, I've made the following changes:

Major refactors:

- Really having to shoehorn in the "subworkflow" idea in the form of treating each module as a builder
- I've tried to leave the ability to run each module by itself.
- Refactoring out `genotype_info` to only run once.
- If the data for a gene in the conditional analysis is constant for all snps, the spearman rank correlation is not defined, and the gene is skipped for the remainder of iterations.

Minor refactors:

- Added checkpoints to more individual functions
- Cleaned-up passing dataframes to python jobs (keep as paths where possible)
- Refactor names to include `_path` where it's a path.
- Refactor some constants as overridable inputs to the workflow with defaults
- Filter early where possible to save compute


KatalinaBobowik (approved):
looks great! I like the extra comments in there as well

scripts/eqtl_hail_batch/launch_eqtl_spearman.py
```diff
@@ -0,0 +1,1501 @@
...
+        --input-files-prefix 'gs://cpg-tob-wgs-test/scrna_seq/grch38_association_files' \
+        --chromosomes '1 2' \
+        --genes B_intermediate
+"""
+import os
+import logging
+
+
+def setup_logger(name):
+    loggers_to_silence = [
```
illusional, line 23: @lgruen, this is the only way I could figure out to properly silence these. Just the `getLogger` didn't work for me.

PR #122, opened 2022-09-13
script to get rare variants for gene-donor outlier pair

annacuomo:


illusional (changes requested):
Hey @annacuomo, it would be good to add some documentation around what this script does, and how to use it. Eg: Making this a runnable script with a CLI (using click) to select specific chromosomes or samples, adding a documentation header at the top to tell people about why this script is useful, etc.

Does this script output anything, or just print things to stdout?

MattWellie:
Main comment would be that this is an exceptionally specific script to be merged into main. Instead, you can copy this whole MT (or a subset thereof, e.g. this sample &/or gene region) to test, and carry out _ad hoc_ investigations in a notebook, then delete the copied data if it's massive. [This script](https://github.com/populationgenomics/analysis-runner/blob/main/scripts/subset_matrix_table.py) can be run on any dataset thanks to [this PR](https://github.com/populationgenomics/analysis-runner/pull/499)

scripts/rv_expression_association/igll5_vep.py
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env python3
+
+import hail as hl
+import numpy as np
+import pandas as pd
+from hail.methods import export_plink
+from cpg_utils.hail_batch import dataset_path, init_batch, output_path #, reference_path
+
+# object containing variants within a 50K window on either side of the IGLL5 gene
+MT = dataset_path('v0/IGLL5_50K_window.mt')
```
illusional: Files tend to be grouped by type, then versioned. I would've have imagined something like `IGLL5/v0/<files>`. Not exactly part of this PR because you're reading it, but just noting.

scripts/rv_expression_association/igll5_vep.py
```diff
@@ -0,0 +1,69 @@
...
+    mt = mt.annotate_rows(vep=vep_ht[mt.row_key].vep)
+
+    print(mt.count())
+
+    # filter rare variants only (MAF < 5%)
+    mt = hl.variant_qc(mt)
+    rv_mt = mt.filter_rows(mt.variant_qc.AF[1] < 0.05)
+    rv_mt = rv_mt.filter_rows(rv_mt.variant_qc.AF[1] > 0)
+
+    print(rv_mt.count())
```
illusional: We tend to prefer the `logging` library for logging stuff to the console, I'd probably encourage labels to go along with these print statements so they're not just numbers in stderr.

scripts/rv_expression_association/prepare_inputs.py
```diff
@@ -0,0 +1,172 @@
...
+# from cloudpathlib import AnyPath
+from cpg_utils import to_path
+from pandas_plink import read_plink1_bin
+from limix.qc import quantile_gaussianize
+
+# use logging to print statements, display at info level
+logging.basicConfig(format='%(levelname)s:%(message)s', level=logging.INFO)
+
+# sample_mapping_file = 'gs://cpg-tob-wgs-test/v0/skat/smf_Bcells.csv'
+phenotype_file = 'gs://cpg-tob-wgs-test/v0/skat/sce22.h5ad'
```
illusional: We prefer to parameterise inputs (as opposed to constants defined in the top) in the function body with `click` for the command-line interface.

illusional (changes requested):
This PR needs more documentation + some cleanup before being merged to main, there are a few TODOs and large blocks of commented-out code. `prepare_inputs.py` could be more descriptive.

I'd recommend a README in the same directory which could suggest why you might run this script, context and an example of how to run it.

scripts/rv_expression_association/igll5_vep.py
annacuomo: Hi Michael, sorry I made this a normal PR instead of a draft one by mistake! That said I think I now got it to work using notebooks / test objects and analysis-runner, so hopefully it's almost ready to review (I have also added a README).

As for the files structure I appreciate that there isn't much of one at the moment, again I was still very much in exploratory / testing phase.. what is the easiest way to re-arrange files? And also best practice when switching to main, presumably we want the same hierarchy if possible?

scripts/rv_expression_association/igll5_vep.py
annacuomo: Yes sorry, should be done now, but how do you deal with this when it's a table / multi-line? i.e. https://github.com/populationgenomics/tob-wgs/blob/get-variants/scripts/rv_expression_association/get_vep_variants.py#L58, which outputs
Screen Shot 2022-09-27 at 10 05 12 am

scripts/rv_expression_association/prepare_inputs.py
annacuomo: on it!

scripts/rv_expression_association/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Overview
+This folder contains three scripts:
+* [get_vep_variants.py](get_vep_variants.py) is a Python script which takes a MT + HT object, filters relevant variants (detail below) and exports to plink files
+* [prepare_inputs.py](prepare_inputs.py) is a Python script which takes in genotype and expression files and prepares the input files to run SKAT
+* [run_SKAT.R](run_SKAT.R) is an R script that runs SKAT using the inputs generated from the previous scripts and returns association p-values
+
+### Step 1 - select genetic variants
+
+At the moment, I have subsetted both the MT object and the [VEP](https://asia.ensembl.org/info/docs/tools/vep/index.html)-annotated object manually, using [this script](https://github.com/populationgenomics/analysis-runner/blob/main/scripts/subset_matrix_table.py) to a specific genomic region around this one specific gene (_VPREB3_), _e.g._, for the VEP-annotated hail table (from local version of populationgenomics/analysis-runner/scripts/):
```
lgruen:
```suggestion
At the moment, I have subset both the MT object and the [VEP](https://asia.ensembl.org/info/docs/tools/vep/index.html)-annotated object manually, using [this script](https://github.com/populationgenomics/analysis-runner/blob/main/scripts/subset_matrix_table.py) to a specific genomic region around this one specific gene (_VPREB3_), _e.g._, for the VEP-annotated hail table (from local version of populationgenomics/analysis-runner/scripts/):
```

scripts/rv_expression_association/README.md
```diff
@@ -0,0 +1,97 @@
...
+```
+analysis-runner --dataset "tob-wgs" \
+    --description "prepare SKAT input files" \
+    --access-level "test" \
+    --output-dir "v0" \
+    prepare_inputs.py
+```
+
+#### current steps
+for now, considering all regulatory variants (or, _e.g.,_ selecting promoter variants only) and considering pseudo-bulk (mean) expression for each individual, borrowing the expression file (here, donor-level aggregated expression in naive B cells) from Kat's eQTL files.
```
lgruen:
```suggestion

For now, considering all regulatory variants (or, _e.g.,_ selecting promoter variants only) and considering pseudo-bulk (mean) expression for each individual, borrowing the expression file (here, donor-level aggregated expression in naive B cells) from Kat's eQTL files.
```

scripts/rv_expression_association/prepare_inputs.py
```diff
@@ -0,0 +1,212 @@
...
+import pandas as pd
+import xarray as xr
+from cloudpathlib import AnyPath
+from cpg_utils import to_path
+from cpg_utils.hail_batch import output_path
+
+subprocess.run([sys.executable, '-m', 'pip', 'install', 'limix==3.0.4'], check=True)
+subprocess.run(
+    [sys.executable, '-m', 'pip', 'install', 'pandas_plink==2.2.9'], check=True
+)
```
lgruen:
```suggestion
subprocess.run([sys.executable, '-m', 'pip', 'install', 'limix==3.0.4', 'pandas_plink==2.2.9'], check=True)
```

scripts/rv_expression_association/prepare_inputs.py
```diff
@@ -0,0 +1,212 @@
...
+
+from pandas_plink import read_plink1_bin  # lint complains about these but surely i need to install them before i can import them?
```
lgruen: You're absolutely right. Maybe consider suppressing the linter warning here.

```suggestion
from pandas_plink import read_plink1_bin
```

scripts/rv_expression_association/prepare_inputs.py
```diff
@@ -0,0 +1,212 @@
...
+    phenotype_file: str,
+    kinship_file: str,
+    output_folder: str,
+):
+
+    expression_filename = AnyPath(output_path(f'{gene_name}_{cell_type}.csv'))
+    genotype_filename = AnyPath(output_path(f'{gene_name}_rare_regulatory.csv'))
+    kinship_filename = AnyPath(output_path('kinship_common_samples.csv'))
+
+    ##################### lint also not happy with these, should I care?
```
lgruen: What do you mean by this? What's the linter complaining about?

scripts/rv_expression_association/prepare_inputs.py
```diff
@@ -0,0 +1,212 @@
...
+
+    ## samples in kinship
+    donors_e_short = [re.sub('.*_', '', donor) for donor in donors_e]
+    donors_k = sorted(set(list(K.sample_0.values)).intersection(donors_e_short))
+
+    ######################################
+    ### subset files to common samples ###
+    ######################################
+
+    ###############
```
lgruen: You might want to consider using [marker regions](https://code.visualstudio.com/docs/editor/codebasics#_folding) instead.

scripts/rv_expression_association/get_vep_variants.py
```diff
@@ -0,0 +1,98 @@
...
+logging.basicConfig(format='%(levelname)s:%(message)s', level=logging.INFO)
+
+
+def main():
+    # read and densify object
+    init_batch()
+    mt = hl.read_matrix_table(MT)
+    mt = hl.experimental.densify(mt)
+
+    logging.info('Number of variants in window: {}'.format(mt.count()[0]))
```
lgruen: `mt.count()` might lead to the expensive `densify` operation being run redundantly here. Consider running `densify` _after_ this print statement instead, to avoid that duplication, as it won't change the row count.

scripts/rv_expression_association/get_vep_variants.py
```diff
@@ -0,0 +1,98 @@
...
+
+    # filter out low quality variants
+    mt = mt.filter_rows(hl.len(hl.or_else(mt.filters, hl.empty_set(hl.tstr))) == 0)
+
+    # consider biallelic variants only (no multi-allelic, no ref-only)
+    mt = mt.filter_rows(hl.len(mt.alleles) == 2)
+    mt = mt.filter_rows(hl.is_snp(mt.alleles[0], mt.alleles[1]))
```
lgruen: Consider combining these filter predicates, as they're not dependent on one another:

```suggestion
    mt = mt.filter_rows(
        # filter out low quality variants
        (hl.len(hl.or_else(mt.filters, hl.empty_set(hl.tstr))) == 0) &
        # consider biallelic variants only (no multi-allelic, no ref-only)
        (hl.len(mt.alleles) == 2) &
        hl.is_snp(mt.alleles[0], mt.alleles[1]))
```

scripts/rv_expression_association/get_vep_variants.py
```diff
@@ -0,0 +1,98 @@
...
+
+    # annotate using VEP
+    vep_ht = hl.read_table(VEP_HT)
+    mt = mt.annotate_rows(vep=vep_ht[mt.row_key].vep)
+
+    logging.info('Number of QC-passing, biallelic SNPs: {}'.format(mt.count()[0]))
+
+    # filter rare variants only (MAF < 5%)
+    mt = hl.variant_qc(mt)
+    rv_mt = mt.filter_rows(mt.variant_qc.AF[1] < 0.05)
```
lgruen: Ditto.

scripts/rv_expression_association/get_vep_variants.py
```diff
@@ -0,0 +1,98 @@
...
+    rv_mt = rv_mt.filter_rows(rv_mt.variant_qc.AF[1] > 0)
+
+    logging.info('Number of rare variants (freq<5%): {}'.format(rv_mt.count()[0]))
```
lgruen: Again, the `count` operation will lead to all the operations being run. Consider using `checkpoint` to avoid that, if you need the count here (and in the log statement below).

scripts/rv_expression_association/get_vep_variants.py
```diff
@@ -0,0 +1,98 @@
...
+    )
+
+    # print out stats
+    # types of regulatory variants
+    biotypes = pd.Series(
+        filtered_mt.vep.regulatory_feature_consequences['biotype'].collect()
+    )
+    print(
+        biotypes.value_counts()
+    )  # not sure how to turn this into logging? Multiline table
```
lgruen: Do you need to log this at all?

lgruen (changes requested):

scripts/rv_expression_association/prepare_inputs.py
annacuomo: I didn't realise I could do that! How? I also have another recurrent lint comment about having "multiple modules named main" which I don't feel I can address really

scripts/rv_expression_association/prepare_inputs.py
annacuomo: "too many leading '#' for block comment"

scripts/rv_expression_association/get_vep_variants.py
annacuomo: will do, thanks!

scripts/rv_expression_association/get_vep_variants.py
annacuomo: I guess at this stage I was interested to see specifically how many variants each filter was removing, as I am still in the process of determining what the best selection strategy would be? For example, counting how many of the freq<5% variants are actually freq<1% (as it turns out, the vast majority) is relevant information that goes against going my original strategy of splitting near-common vs actually rare variants?

scripts/rv_expression_association/get_vep_variants.py
annacuomo: Maybe I could save these stats (instead of printing them out), if the count() operation is especialy expensive?

scripts/rv_expression_association/get_vep_variants.py
annacuomo: same as above, some biotypes are more likely regulatory than others, so useful to explore this in a few cases to inform my strategy - but again I could just save this information in a table instead of printing these logs, if they are very expensive

scripts/rv_expression_association/README.md
```diff
@@ -0,0 +1,98 @@
+## Analysis plan
+
+The overall goal of these scripts is to run a gene-set association test to test for an association between i) a number of rare genetic variants located in and around a gene and ii) the expression level of the gene itself (in a given cell type).
```
KatalinaBobowik, line 3: This runs a gene-set association analysis to test for an association between i) the number of rare genetic variants located in and 50kb up or down-stream of the gene and ii) the expression level of the gene itself (in a given cell type).

KatalinaBobowik:
```suggestion
This runs a gene-set association analysis to test for an association between i) the number of rare genetic variants located in and 50kb up or down-stream of the gene and ii) the expression level of the gene itself (in a given cell type).
```

scripts/rv_expression_association/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+At first, I am testing specifically one chromosome 22 gene (_VPREB3_), in naive B cells.
```
KatalinaBobowik:
```suggestion
Testing is done on one gene within chromosome 22, VPREB3, in naive B cells, which will be further expanded to multiple genes.
```

KatalinaBobowik: Out of curiosity, is this gene randomly-chosen?

scripts/rv_expression_association/README.md
```diff
@@ -0,0 +1,98 @@
...
+I consider variants located at most 50kb up or down-stream of the gene body (including those within the gene itself).
+Later, I'll want to run this for many genes.
+
+### Overview
+This folder contains three scripts:
+* [get_vep_variants.py](get_vep_variants.py) is a Python script which takes a MT + HT object, filters relevant variants (detail below) and exports to plink files
```
KatalinaBobowik:
```suggestion
* [get_vep_variants.py](get_vep_variants.py) is a Python script which takes a hail MT + HT object, filters for relevant variants (detailed below), and exports to plink files
```

scripts/rv_expression_association/run_SKAT.R
```diff
@@ -0,0 +1,100 @@
...
+# set bucket
+googleCloudStorageR::gcs_global_bucket("gs://cpg-tob-wgs-test")
+
+# sample mapping file, matching CPG IDs to OneK1K IDs
+smf <- googleCloudStorageR::gcs_get_object("scrna-seq/grch38_association_files/OneK1K_CPG_IDs.tsv")
+smf_df <- as.data.frame(smf)
+colnames(smf_df)[2] <- "CPG_ID"  # change column name to match other files
+
+# expression file (gene: VPREB3, cell type: naive B cells)
+y <- googleCloudStorageR::gcs_get_object("v0/vpreb3_B_naive_expression.csv")
```
KatalinaBobowik: Just a general nit comment, but although I think these are the variable names SKAT uses, they're quite hard to follow

scripts/rv_expression_association/run_SKAT.R
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env Rscript
+
```
KatalinaBobowik, line 7: Perhaps here you could just have a one-liner about the main purpose of the script

scripts/rv_expression_association/README.md
annacuomo: A mix! I had selected chromosome 22 somewhat at random (I was under the impression that this was the shortest chromosome, turns out that is actually chrom 21), but then since I was looking at expression in naive B cells I selected genes that were differentially expressed between naive B cells and all other cell types, hoping it would be an indication of a gene that is relevant for that specific cell type. That said at this stage this is largely just to start somewhere to build a working pipeline :)

scripts/rv_expression_association/run_SKAT.R
```diff
@@ -0,0 +1,100 @@
...
+suppressPackageStartupMessages(suppressMessages(install.packages("googleCloudStorageR", repos = 'http://cran.csiro.au', quietly = T)))
```
KatalinaBobowik: Just curious if you actually need these suppression messages, or if it just makes the logs look cleaner? If it's important, maybe add a general comment about why you need it

scripts/rv_expression_association/run_SKAT.R
annacuomo: ok, yes as you said I was trying to stick with the SKAT nomenclature but will try to give more informative names! As a general comment while I use R a lot for plotting I have not really written scripts like this very often (vs Python), so apologies for the poor form

scripts/rv_expression_association/run_SKAT.R
annacuomo: will do, thanks!

scripts/rv_expression_association/run_SKAT.R
annacuomo: no i don't at all (it's just aesthetics), and in fact this didn't even really work! I might drop them for now and if I do find a working solution add a comment to clarify this is just to make the log cleaner

scripts/rv_expression_association/prepare_inputs.py
lgruen: You can add comments to the offending lines, e.g.

```
# pylint: disable=locally-disabled
```

Where `locally-disabled` is the name of the warning you'd like to suppress.

scripts/rv_expression_association/prepare_inputs.py
lgruen: Maybe just switch to regions in case you like those!

scripts/rv_expression_association/get_vep_variants.py
lgruen: Oh, right, I missed that you're using _both_ tables below, sorry!

scripts/rv_expression_association/get_vep_variants.py
lgruen: `count` itself isn't particularly expensive -- what's expensive is that it triggers the evaluation of the tree of operations that are lazily executed in Hail. The results of that are _not_ automatically cached. E.g. consider this example:

```py
mt = hl.read_matrix_table(...)
mt = mt.filter_rows(...)
print(mt.count())
print(mt.count())
```

This will run `filter_rows` _twice_, which might be counterintuitive at first, but is a result of the lazy evaluation of expressions: https://hail.is/docs/0.2/overview/expressions.html

Instead, you can rewrite this as follows:

```py
mt = hl.read_matrix_table(...)
mt = mt.filter_rows(...)
mt = mt.checkpoint(...)
print(mt.count())
print(mt.count())
```

This explicitly writes out the intermediate result (which will use storage, but the `tmp` bucket is a good destination for this), avoiding the repeated evaluation of `filter_rows` in order to evaluate `count`.

scripts/rv_expression_association/get_vep_variants.py
lgruen: (See my comment above -- it's not the logs itself (although a table might be more handy for post-processing in a notebook), but the repeated evaluation.)

lgruen (changes requested):
(Added some more comments as responses to existing threads.)

scripts/rv_expression_association/prepare_inputs.py
annacuomo: i am so sorry, I don't get why what I just changed doesn't work?

scripts/rv_expression_association/prepare_inputs.py
annacuomo: i like the idea but am finding the VSCode documentation a bit obscure still :( do you have a good example in python by any chance? Sorry for being slow

scripts/rv_expression_association/get_vep_variants.py
annacuomo: ok, i've tried to add checkpoints, maybe check if i missed any?

scripts/rv_expression_association/prepare_inputs.py
lgruen: I think it did work -- the remaining linter errors are about the *order* of other import statements, not the *position*:

```
scripts/rv_expression_association/prepare_inputs.py:4:0: C0411: standard import "import logging" should be placed before "import click" (wrong-import-order)
scripts/rv_expression_association/prepare_inputs.py:5:0: C0411: standard import "import re" should be placed before "import click" (wrong-import-order)
scripts/rv_expression_association/prepare_inputs.py:6:0: C0411: standard import "import subprocess" should be placed before "import click" (wrong-import-order)
scripts/rv_expression_association/prepare_inputs.py:7:0: C0411: standard import "import sys" should be placed before "import click" (wrong-import-order)
```

annacuomo: mmmm am I looking at the wrong thing?

https://github.com/populationgenomics/tob-wgs/actions/runs/3141000124/jobs/5102974062

scripts/rv_expression_association/prepare_inputs.py
lgruen: It's fairly simple to use -- here's an example for Python: https://github.com/populationgenomics/tob-wgs/blob/7fddfcc795672f5d6f66e43ce9bec12467412119/scripts/eqtl_hail_batch/launch_eqtl_spearman.py#L75

annacuomo: sorry, perfect, thanks!

scripts/rv_expression_association/prepare_inputs.py
lgruen: The report in that link doesn't contain any warning for `wrong-import-position`, which is the category you've disabled. So I think this is working.

annacuomo: sorry, I think my confusion comes from the fact that while pylint doesn't flag that problem anymore, flake8 still does? I didn't realise those two may sometime be overlapping, and couldn't understand why the problem was still there

lgruen: Oh, right -- there's unfortunately significant overlap between pylint and flake8. @illusional / @vladsavelyev Any thoughts on whether we should just pick one of them instead of running both?

scripts/rv_expression_association/get_vep_variants.py
```diff
@@ -0,0 +1,94 @@
...
+    rv_mt_path = output_path('rare_variants.mt', 'tmp')
+    rv_mt = rv_mt.checkpoint(rv_mt_path, overwrite=True)
+    logging.info(f'Number of rare variants (freq<5%): {rv_mt.count()[0]}')
+
+    # filter variants found to have regulatory effects
+    filtered_mt = rv_mt.filter_rows(
+        hl.len(rv_mt.vep.regulatory_feature_consequences['biotype']) > 0
+    )
+    logging.info(
+        f'Number of rare variants (freq<5%) with ergulatory conequences: {filtered_mt.count()[0]}'
```
lgruen:
```suggestion
        f'Number of rare variants (freq<5%) with regulatory consequences: {filtered_mt.count()[0]}'
```

scripts/rv_expression_association/get_vep_variants.py
```diff
@@ -0,0 +1,94 @@
...
+    )
+
+    filtered_rrv_mt = filtered_mt.filter_rows(filtered_mt.variant_qc.AF[1] < 0.01)
+    logging.info(f'Number of rarer variants (freq<1%): {filtered_rrv_mt.count()[0]}')
+
+    filtered_0maf_mt = filtered_mt.filter_rows(filtered_mt.variant_qc.AF[1] == 0)
+    logging.info(
+        f'Check that there are {filtered_0maf_mt.count()[0]} variants with freq==0'
+    )
```
lgruen: Just double-checking that you really just want the counts in the log and not store the resulting table for further downstream processing somewhere?

scripts/rv_expression_association/get_vep_variants.py
```diff
@@ -0,0 +1,94 @@
...
+    print(
+        biotypes.value_counts()
+    )  # not sure how to turn this into logging? Multiline table
+
+    # MAF distribution
+    mafs = pd.Series(filtered_mt.variant_qc.AF[1].collect())
+    logging.info(f'Min frequency in set: {np.nanmin(mafs)}')
+    logging.info(f'Max frequency in set: {np.nanmax(mafs)}')
+    logging.info(f'Mean frequency in set: {np.nanmean(mafs)}')
```
lgruen: `collect` doesn't scale well, as it'll require a lot of memory on a single node. Could you use Hail's `stats` method here instead?
https://hail.is/docs/0.2/aggregators.html#hail.expr.aggregators.stats

scripts/rv_expression_association/get_vep_variants.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # export MT object to PLINK (all regulatory variants)
+    filtered_mt_path = output_path('plink_files/vpreb3_rare_regulatory')
+    export_plink(filtered_mt, filtered_mt_path, ind_id=filtered_mt.s)
```
lgruen, line 66: This will rerun the `filter_rows` from line 49, which already ran before due to logging the row count.

scripts/rv_expression_association/get_vep_variants.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # filter to promoter variants only
+    filtered_mt2 = filtered_mt.filter_rows(
+        filtered_mt.vep.regulatory_feature_consequences['biotype'][0] == 'promoter'
+    )
+    print(filtered_mt2.count())
+    # export to PLINK
+    filtered_mt_path = output_path('plink_files/vpreb3_rare_promoter')
+    export_plink(filtered_mt2, filtered_mt_path, ind_id=filtered_mt2.s)
```
lgruen: Ditto.

lgruen (changes requested):

scripts/rv_expression_association/get_vep_variants.py
lgruen: You got almost all of them :). Flagged two more spots.

scripts/rv_expression_association/prepare_inputs.py
MattWellie: The latest template includes flake8, mypy, black, and pylint. Maybe reducing that to pylint and black would be sufficient?

scripts/rv_expression_association/get_vep_variants.py
annacuomo: great, will definitely use Hail's `stats` method to get min, max and mean! However if I wanted to save info in a table like i mentioned before, i was going to use to `collect` to get maf values, add them to a pandas data frame and then save as csv? Are you suggesting to use a Hail `table` instead?

scripts/rv_expression_association/get_vep_variants.py
annacuomo: yes i think i will go the table route (writing it now) and will drop these lines

scripts/rv_expression_association/get_vep_variants.py
annacuomo: ok! I think I got it now, but still mildly confusing so forgive me if I still missed some

scripts/rv_expression_association/prepare_inputs.py
annacuomo: in the meantime, is there an easy way to disable flake8 for those same lines?

scripts/rv_expression_association/get_vep_variants.py
annacuomo: follow up, i tried to save as a hail table, maybe check syntax? then, i can delete commented out lines

lgruen (changes requested):
Did something with a recent merge go wrong? This PR now includes unrelated files like `scripts/eqtl_hail_batch/launch_eqtl_spearman.py`...

scripts/rv_expression_association/get_vep_variants.py
```diff
@@ -0,0 +1,104 @@
...
+    filtered_mt_path = output_path('regulatory_variants.mt', 'tmp')
+    filtered_mt = filtered_mt.checkpoint(filtered_mt_path, overwrite=True)  # checkpoint
+    logging.info(
+        f'Number of rare variants (freq<5%) with regulatory consequences: {filtered_mt.count()[0]}'
+    )
+
+    # export this as a Hail table for downstream analysis
+    ht_filename = output_path('vpreb3_rare_regulatory_summary.ht')
+    ht = filtered_mt.rows()
+    ht.write(ht_filename)  # will this work? should i select important colunmns only?
```
lgruen: By "columns", do you mean fields? You've already selected `.rows()` above, so this will only include any row fields. You can use `.select()` if you want to further restrict the fields that are written.

scripts/rv_expression_association/get_vep_variants.py
lgruen: Did you mean to add another checkpoint for `filtered_mt2` here?

scripts/rv_expression_association/get_vep_variants.py
lgruen: Looks like you're now saving the full table (I've added comments there, too).

But coming back to your previous question: you can easily serialize the `struct` that https://hail.is/docs/0.2/aggregators.html#hail.expr.aggregators.stats returns, e.g. to JSON, in case you need to save that for downstream analyses.

lgruen (changes requested):

MattWellie:
@lgruen that's my bad. I was trying to rebase to adopt existing main branch changes that were causing MyPy to fail. There was a change during the rebase, so it became a rebase-merge-rebase, and now github is stuggling to reconcile the changes. #131 is a context PR which is cleaner WRT `main`

Just to check the content is the same:
https://github.com/populationgenomics/tob-wgs/compare/get-variants...checkpoint_post_matt?expand=1

lgruen:
> @lgruen that's my bad. I was trying to rebase to adopt existing main branch changes that were causing MyPy to fail. There was a change during the rebase, so it became a rebase-merge-rebase, and now github is stuggling to reconcile the changes. #131 is a context PR which is cleaner WRT `main`
> 
> Just to check the content is the same: https://github.com/populationgenomics/tob-wgs/compare/get-variants...checkpoint_post_matt?expand=1

Ah, thanks! Do you maybe want to `reset` back to [34be8c6](https://github.com/populationgenomics/tob-wgs/pull/122/commits/34be8c6ac2809a22577ad8601d27e3a9b325944f) and merge `main`, followed by a force-push to clean things up? Rebasing is always a bit tricky with such a large PR and concurrent changes.

MattWellie:
> Ah, thanks! Do you maybe want to `reset` back to [34be8c6](https://github.com/populationgenomics/tob-wgs/pull/122/commits/34be8c6ac2809a22577ad8601d27e3a9b325944f) and merge `main`, followed by a force-push to clean things up?

I've done this, and I hate the results. I don't feel particularly comfortable with the current state (the diff has lots of deletions in Kat's scripts) so I'm planning on resetting to [34be8c6](https://github.com/populationgenomics/tob-wgs/pull/122/commits/34be8c6ac2809a22577ad8601d27e3a9b325944f) , force-pushing, and pretending I was never here

scripts/rv_expression_association/get_vep_variants.py
annacuomo: yes sorry i meant fields I guess! I am happy for this to be about variants only (so doing `.rows()` should be correct here ), but at the moment I am essentially only looking at allele frequencies and VEP annotations, so wondering whether to select those fields only, or whether it's safer to keep everything in case i have e.g., QC concerns in the future! is there a way to drop fields that are NAs everywhere? For example, I believe the `rsid` field was empty. Are there other consideration on e.g. memory that are relevant here?

scripts/rv_expression_association/get_vep_variants.py
annacuomo: I think if I save the whole table these stats should be easy to pull out when i need them in the future? E.g. I might select variants that vep annotated to be promoter, and check the allele frequency distribution for those variants specifically, etc?

scripts/rv_expression_association/prepare_inputs.py
annacuomo: sorry, does anyone have ideas re: flake8?

annacuomo: and what about a way to fix the mypy error? 

```
> mypy.....................................................................Failed
> - hook id: mypy
> - exit code: 2
> 
> scripts/vep/main.py: error: Duplicate module named "main" (also at "scripts/generate_genotype_df/main.py")
> scripts/vep/main.py: note: Are you missing an __init__.py? Alternatively, consider using --exclude to avoid checking one of them.
> Found 1 error in 1 file (errors prevented further checking)
```

I think this is what Matt was trying to do, this has apparently been fixed elsewhere, but I am working off an older version? Sorry in advance if I am misunderstanding

scripts/rv_expression_association/get_vep_variants.py
annacuomo: should I? I am printing anything anymore, but will is still do the filtering twice before exporting to plink if I don't add a checkpoint?

scripts/rv_expression_association/get_vep_variants.py
lgruen: It should be fine to just store the whole `.rows` table. Typically it's the large number of samples that really blow up the table sizes.

scripts/rv_expression_association/get_vep_variants.py
lgruen: Yup, makes sense!

scripts/rv_expression_association/get_vep_variants.py
lgruen: Did you mean you're _not_ printing anything anymore? Line 87 still says `print(filtered_mt2.count())`, so that line and `export_plink(filtered_mt2, filtered_mt_path, ind_id=filtered_mt2.s)` will both cause the `filter_rows` in line 84 to be executed.

lgruen (changes requested):

scripts/rv_expression_association/prepare_inputs.py
lgruen: Did you see @MattWellie changes in https://github.com/populationgenomics/tob-wgs/pull/131/files to `.flake8`? Does that help?

MattWellie: AFAIK this MyPy error will go away on main, as those two files have already been renamed there...

scripts/rv_expression_association/get_vep_variants.py
annacuomo: ok!

scripts/rv_expression_association/prepare_inputs.py
annacuomo: ok perfect!

scripts/rv_expression_association/get_vep_variants.py
annacuomo: should be all done now? The other comment you have is outdated I think, I did delete that print statement in my last push

scripts/rv_expression_association/prepare_inputs.py
annacuomo: > Did you see @MattWellie changes in https://github.com/populationgenomics/tob-wgs/pull/131/files to `.flake8`? Does that help?

does this need to be approved?

annacuomo: > AFAIK this MyPy error will go away on main, as those two files have already been renamed there...

ok great!

MattWellie: @annacuomo  I think #131 should be scrapped, best just to copy that one-line change into your branch

annacuomo: > @annacuomo I think #131 should be scrapped, best just to copy that one-line change into your branch

aaaah I see! Sorry for being slow, should've done now

annacuomo: ok i think i'm down to only the mypy error, which hopefully should be resolved already in main, is that right @lgruen ? Any more concerns?

MattWellie (dismissed):

annacuomo:
> Hey @annacuomo, it would be good to add some documentation around what this script does, and how to use it. Eg: Making this a runnable script with a CLI (using click) to select specific chromosomes or samples, adding a documentation header at the top to tell people about why this script is useful, etc.
> 
> Does this script output anything, or just print things to stdout?

@illusional A lot of things changed since this comment, is it better now?

scripts/rv_expression_association/prepare_inputs.py
```diff
@@ -0,0 +1,211 @@
...
+        'limix==3.0.4',
+        'pandas_plink==2.2.9',
+        'xarray==0.20.2',
+    ],
+    check=True,
+)
+
+import xarray as xr  # pylint: disable=wrong-import-position, import-error
+from pandas_plink import (read_plink1_bin)  # pylint: disable=wrong-import-position, import-error
+from limix.qc import (quantile_gaussianize)  # pylint: disable=wrong-import-position, import-error
```
MattWellie:
```suggestion
# pylint: disable=wrong-import-position,import-error
import xarray as xr  
from pandas_plink import read_plink1_bin
from limix.qc import quantile_gaussianize
```

`A disable comment can be inserted on its own line, applying the disable to everything that comes after in the same block. Alternatively, it can be inserted at the end of the line for which it is meant to apply.`

scripts/rv_expression_association/prepare_inputs.py
```diff
@@ -0,0 +1,211 @@
...
+    # endregion KINSHIP_FILE
+
+    # region GENOTYPE_FILE
+
+    # read in genotype file (plink format)
+    # bed
+    with to_path(genotype_file_bed).open('rb') as handle:
+        data = handle.readlines()
+    with open('temp.bed', 'wb') as handle:
+        handle.writelines(data)
```
MattWellie:
```suggestion
    with to_path(genotype_file_bed).open('rb') as read_handle:
        with open('temp.bed', 'wb') as write_handle:
            write_handle.writelines(read_handle.readlines())
```

This is a little fiddly, but doesn't require any intermediate object to hold the file content

scripts/rv_expression_association/prepare_inputs.py
```diff
@@ -0,0 +1,211 @@
...
+    # bim
+    with to_path(genotype_file_bim).open('rb') as handle:
+        data = handle.readlines()
+    with open('temp.bim', 'wb') as handle:
+        handle.writelines(data)
```
MattWellie:
```suggestion
    with to_path(genotype_file_bim).open('rb') as read_handle:
        with open('temp.bim', 'wb') as write_handle:
            write_handle.writelines(read_handle.readlines())
```

scripts/rv_expression_association/prepare_inputs.py
```diff
@@ -0,0 +1,211 @@
...
+    # fam
+    with to_path(genotype_file_fam).open('rb') as handle:
+        data = handle.readlines()
+    with open('temp.fam', 'wb') as handle:
+        handle.writelines(data)
```
MattWellie:
```suggestion
    with to_path(genotype_file_fam).open('rb') as read_handle:
        with open('temp.fam', 'wb') as write_handle:
            write_handle.writelines(read_handle.readlines())
```

scripts/rv_expression_association/prepare_inputs.py
```diff
@@ -0,0 +1,211 @@
...
+    # this file will map different IDs (and OneK1K ID to CPG ID)
+    sample_mapping = pd.read_csv(sample_mapping_file, sep='\t')
+
+    # samples with expression data
+    donors_onek1k = sample_mapping['OneK1K_ID'].unique()
+    donors_onek1k.sort()
+    donors_exprs = sorted(
+        set(list(phenotype.sample.values)).intersection(donors_onek1k)
+    )
+    logging.info(f'Number of unique donors with expression data: {len(donors_exprs)}')
```
MattWellie:
```suggestion
    # samples with expression data
    donors_exprs = set(phenotype.sample.values).intersection(
        set(sample_mapping['OneK1K_ID'].unique())
    )
    
    logging.info(f'Number of unique donors with expression data: {len(donors_exprs)}')
```

I think there are some extra steps here that aren't helping. `unique()` returns a series, which can be cast directly as a list/set.

This is being used downstream in a set intersection, so the ordering doesn't matter, in which case set operations are faster, and you can skip a few steps.

scripts/rv_expression_association/prepare_inputs.py
```diff
@@ -0,0 +1,211 @@
...
+
+    # samples with genotype data
+    donors_cpg = sample_mapping['InternalID'].unique()
+    donors_cpg.sort()
+    donors_geno = sorted(set(list(geno.sample.values)).intersection(donors_cpg))
+    logging.info(f'Number of unique donors with genotype data: {len(donors_geno)}')
```
MattWellie:
```suggestion
    # samples with genotype data
    donors_geno = set(geno.sample.values).intersection(
        set(sample_mapping['InternalID'].unique())
    )
    logging.info(f'Number of unique donors with genotype data: {len(donors_geno)}')
```

similar story here

MattWellie (commented):
added various comments

scripts/rv_expression_association/prepare_inputs.py
annacuomo: Right! I was going for option 2 here, should I just add it once at the top instead? I thought it made sense to only apply to relevant lines but maybe all the repetition is not great either?

scripts/rv_expression_association/prepare_inputs.py
annacuomo: ok, as long as it works! :)

scripts/rv_expression_association/prepare_inputs.py
annacuomo: i can't remember fully, but i think there was a reason i was sorting them for later? I agree the first `.sort()` is useless, but would sorting the final set make sense? i.e. `sorted(set(...))`

scripts/rv_expression_association/prepare_inputs.py
illusional: Why not just copy this to disk?

```python
to_path(genotype_file_bed).copy('temp.bed')

File: scripts/rv_expression_association/prepare_inputs.py [hunk 33]
```diff
@@ -0,0 +1,209 @@
+#!/usr/bin/env python3  # pylint: disable=missing-module-docstring
```
illusional: Put pylint after the shebang:

```suggestion
#!/usr/bin/env python3
# pylint: disable=missing-module-docstring,wrong-import-position,import-error
```

scripts/rv_expression_association/prepare_inputs.py
MattWellie: Looks great, didn't know that method existed
https://cloudpathlib.drivendata.org/~latest/api-reference/cloudpath/#cloudpathlib.cloudpath.CloudPath.copy

illusional (commented):
Documentation looks really great! I'm not confident in reviewing the analysis structure, but rest looks good to me 👍

scripts/rv_expression_association/prepare_inputs.py
annacuomo: perfect thank you!!

scripts/rv_expression_association/prepare_inputs.py
MattWellie: In this instance you're applying the same 2 disables to all 3 lines, and the disable only relates to imports, so it's pretty safe to include it once

scripts/rv_expression_association/prepare_inputs.py
MattWellie: `sorted` turns the set into a list. Later this is used in the `is_in()` operation, where a set would be faster, and the ordering won't alter the results.

e.g. as far as I'm aware when you do

```
sample_mapping.loc[sample_mapping['OneK1K_ID'].isin(donors_exprs)]
```

the ordering of the results is based on `sample_mapping`, not `donors_exprs`

annacuomo: ok! Ideally I'd like to return all objects (y, geno, and kinship) not only with the same samples, but in the same order. This is complicated (for my limited abilities) by the fact that the samples are called differently in each of the objects, so arguably I should just change them all to the same? Now i deal with this very uglily (and hopefully correct???) in R, but a better solution would be more than welcome

MattWellie: Aye, definitely makes sense to maintain the ordering where possible. The issue here would be that the objects you're sorting don't ultimately alter the order of the DataFrame. 

I try and stay away from R where possible :D 

It should be simple in Pandas to make sure the ID column is always titled `CPG_ID` then sort the DF on those values to make sure it's consistent across objects

annacuomo: sorry I misspoke - it is not that the column name is different, but rather that the IDs themselves are different - in the expression file, the OneK1K_IDs are used (e.g., 943_944), in the genotype the CPG ones are (e.g., CPG9951)

PR #123, opened 2022-09-15
Rename script to fix mypy linting

illusional:
Mypy has a really awkward error when two files with the same name exist, so just rename this file.


KatalinaBobowik:
Looks good to me

KatalinaBobowik (approved):
👍

PR #125, opened 2022-09-23
Copy chr22 to test

annacuomo:
i would like to copy these input files (from main to test) to test running SKAT on a small dataset, and assess 1) the importance of using a kinship, 2) whether is best to use plink files or can just provide a genotype matrix.

I am testing the gene IGLL5 which is on chrom22, hence the choice of input files to copy, in the future chrom 20 or 21 would be better choices


scripts/data_transfer/copy_chr22_skat_input.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -ex
+
+gsutil -m mv gs://cpg-tob-wgs-main/CellRegMap_input_files/plink_files/plink_chr22.bed gs://cpg-tob-wgs-test/v0/skat/
```
MattWellie: do these need to be moves from main to test, or copies?

MattWellie: AFAIK the usage of -m is different, the syntax would be:

```
gsutil -m mv File File File Target
```

here each file is referenced in a different command, so they can't be copied in parallel

annacuomo: i tried to address this, please check?

annacuomo: copied, good catch!!!

scripts/data_transfer/copy_chr22_skat_input.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -ex
+
+gsutil -m cp gs://cpg-tob-wgs-main/CellRegMap_input_files/plink_files/plink_chr22.bed \
+    gs://cpg-tob-wgs-main/CellRegMap_input_files/plink_files/plink_chr22.bim \
+    gs://cpg-tob-wgs-main/CellRegMap_input_files/plink_files/plink_chr22.fam \
+    gs://cpg-tob-wgs-main/CellRegMap_input_files/GRM/grm_wide.csv \
+    gs://cpg-tob-wgs-main/CellRegMap_input_files/expression_objects/sce22.h5ad gs://cpg-tob-wgs-test/v0/skat/
```
lgruen, line 9: We usually group things by folder, followed by version -- did you mean `skat/v0` instead of `v0/skat`?

scripts/rv_expression_association/igll5_vep.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python3
+
+import hail as hl
+from hail.methods import export_plink
+from cpg_utils.hail_batch import dataset_path, init_batch#, reference_path
```
lgruen:
```suggestion
from cpg_utils.hail_batch import dataset_path, init_batch
```

lgruen (changes requested):
Did you mean to also include the `igll5_vep.py` and `run_SKAT.R` files in this PR? I'm asking because it's generally preferable to keep PRs as small and self-contained as possible. If you want to provide an example of how your data copy can be used (which is great as motivation), you can simply link to another PR in the description.

scripts/data_transfer/copy_chr22_skat_input.sh
illusional: Adding onto this, should they appear in similar places on main / test?

PR #129, opened 2022-09-28
Limit gene-level job parallelism to avoid resource starvation

lgruen:
Context: https://centrepopgen.slack.com/archives/C01FBL6NWFN/p1664243878280569?thread_ts=1664149768.063269&cid=C01FBL6NWFN

### DIFF
--- a/scripts/eqtl_hail_batch/launch_eqtl_spearman.py
+++ b/scripts/eqtl_hail_batch/launch_eqtl_spearman.py
@@ -1404,9 +1404,9 @@ def from_cli(
     chromosomes: str,
     gene_level_parallelism,
     cell_types: list[str] | None,
-    force: bool = False,
-    local_debug: bool = False,
-    gene: list[str] = None,
+    gene: list[str] | None,
+    force: bool,
+    local_debug: bool,
 ):
     """Run the EQTL analysis from command line arguments"""
     chromosomes_list = chromosomes.split(' ') if chromosomes else None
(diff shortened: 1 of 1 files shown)


KatalinaBobowik (approved):
👍 (hopefully that does the trick)

scripts/eqtl_hail_batch/launch_eqtl_spearman.py
```diff
@@ -71,6 +71,11 @@ def setup_logger(name):
 
 MULTIPY_IMAGE = 'australia-southeast1-docker.pkg.dev/cpg-common/images/multipy:0.16'
 
+# Limit parallelism to avoid starting too many jobs near the root of the scheduling
+# tree, leaving too few resources for leaf nodes (particularly Hail Query jobs).  This
+# value is fairly small, as we also process chromosomes and cell types in parallel.
+GENE_LEVEL_PARALLELISM = 20
```
illusional: 20 feels too small, if there are hundreds to thousands of genes per chromosome, this could take a long time

lgruen: I've actually reduced that even further now: 22 chromosomes * 10 genes in parallel * 20 cell types = 4400, and that’s just at the highest level of the scheduling tree. That means we'll use more than 10k cores at any given time. Each of these jobs shouldn't take a lot of time, so I really hope this will actually speed things up.

lgruen: I've added a parameter for this. So in case this turns out to be a poor default setting, we can easily relaunch / tweak it.

PR #130, opened 2022-09-28
Pre annotation filtering + fix previous query order

illusional:
- Minor cleanup of `get_association_effect_data`
- Filter on computed snp_id to avoid annotating whole table + save memory
- Reorder `filter_cols` from #127 to preserve _filtering on rows with a certain variance_ (_outlier sample that has high variance, that SNP will be retained, but if you filter out that sample you’ll have a different subset of variants_) @KatalinaBobowik.

I've checked, and results are identical again - but now using less memory. I think this is ready to test on full scale `chr7:CNTNAP2`.

### DIFF
--- a/scripts/eqtl_hail_batch/launch_eqtl_spearman.py
+++ b/scripts/eqtl_hail_batch/launch_eqtl_spearman.py
@@ -575,6 +575,7 @@ def run_spearman_correlation_scatter(
     mt = hl.filter_intervals(
         mt, [hl.parse_locus_interval(first_and_last_snp, reference_genome='GRCh38')]
     )
+    mt = mt.filter_cols(samples_to_keep.contains(mt['onek1k_id']))
     # remove SNPs with no variance
     mt = mt.filter_rows(
         (hl.agg.all(mt.GT.n_alt_alleles() == 0))
@@ -582,7 +583,6 @@ def run_spearman_correlation_scatter(
         | (hl.agg.all(mt.GT.n_alt_alleles() == 2)),
         keep=False,
     )
-    mt = mt.filter_cols(samples_to_keep.contains(mt['onek1k_id']))
     mt = mt.persist('MEMORY_AND_DISK')
 
     position_table = mt.rows().select()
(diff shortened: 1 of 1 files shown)


scripts/eqtl_hail_batch/launch_eqtl_spearman.py
```diff
@@ -579,15 +579,14 @@ def run_spearman_correlation_scatter(
...
     # remove SNPs with no variance
     mt = mt.filter_rows(
         (hl.agg.all(mt.GT.n_alt_alleles() == 0))
         | (hl.agg.all(mt.GT.n_alt_alleles() == 1))
         | (hl.agg.all(mt.GT.n_alt_alleles() == 2)),
         keep=False,
     )
-    mt = mt.filter_cols(samples_to_keep.contains(mt['onek1k_id']))
-    # TODO: mfranklin to check mt.persist, as mt.rows() is called twice
     mt = mt.persist('MEMORY_AND_DISK')
```
lgruen: `persist` is a no-op in QoB. Did you mean to use `checkpoint`?

lgruen (changes requested):

scripts/eqtl_hail_batch/launch_eqtl_spearman.py
```diff
@@ -579,16 +579,18 @@ def run_spearman_correlation_scatter(
...
         (hl.agg.all(mt.GT.n_alt_alleles() == 0))
         | (hl.agg.all(mt.GT.n_alt_alleles() == 1))
         | (hl.agg.all(mt.GT.n_alt_alleles() == 2)),
         keep=False,
     )
-    mt = mt.filter_cols(samples_to_keep.contains(mt['onek1k_id']))
-    # TODO: mfranklin to check mt.persist, as mt.rows() is called twice
-    mt = mt.persist('MEMORY_AND_DISK')
+    mt = mt.checkpoint(
+        output_path(f'eqtl/{cell_type}/{chromosome}/{gene_name}.mt', 'tmp'),
```
lgruen:
```suggestion
        output_path(f'eqtl/{cell_type}/{chromosome}/{gene_name}/spearman_correlation_scatter_filter.mt', 'tmp'),
```

lgruen (approved):

KatalinaBobowik (approved):
👍

PR #134, opened 2022-09-29
Decide call rate

KatalinaBobowik:
While the 0.8 call rate was used in gnomAD, I want to test whether this can be bumped up. @illusional and I ran into some issues on `test`, where after multiple rounds of conditioning, residual values turn to zero and the gene needs to be dropped out. While this isn't the whole dataset (only 20 in `test`), there can be cases where the top eSNP (the SNP with the most significant p-value) is not chosen for biological reasons, but rather that there are far less samples due to no GT called. As an extreme example, consider a variant which only has a call for 3 individuals with the n alt alleles as 0, 1, and 2. This will result in a perfect correlation, and an extremely high p-value, but only due to technical reasons. Here, I want to test experimentally (on the entire dataset) what the call rate threshold should be by plotting the shape of the call rate (i.e., how many variants have each call rate), the expectation being that a very small number of variants will have a low call rate (in which case, we won't be cutting out much data). 

@lgruen let me know if there would be a better place to put this validation script


scripts/validation/tob_wgs_call_rate_distribution.py
```diff
@@ -0,0 +1,36 @@
...
+    samples = mt.s.collect()
+    n_samples = len(samples)
+    mt = hl.experimental.densify(mt)
+    # filter to biallelic loci only
+    mt = mt.filter_rows(hl.len(mt.alleles) == 2)
+    # filter out variants that didn't pass the VQSR filter
+    mt = mt.filter_rows(hl.len(hl.or_else(mt.filters, hl.empty_set(hl.tstr))) == 0)
+    # Calculate call rate and add to mt
+    samples = mt.s.collect()
+    n_samples = len(samples)
```
lgruen: That's already computed in line 17/18.
```suggestion
```

scripts/validation/tob_wgs_call_rate_distribution.py
```diff
@@ -0,0 +1,36 @@
...
+    mt = mt.annotate_rows(call_rate=1-(hl.agg.sum(hl.is_missing(mt.GT))/n_samples))
```
lgruen:
```suggestion
    mt = mt.annotate_rows(call_rate=hl.agg.sum(hl.is_defined(mt.GT)) / n_samples))
```

scripts/validation/tob_wgs_call_rate_distribution.py
```diff
@@ -0,0 +1,36 @@
...
+    # plot distribution
+    p = hl.plot.histogram(mt.call_rate, range=(0, 1), bins=30, legend='Call Rate', title='Call Rate Distribution')
```
lgruen: Are those enough bins?

Would a `cumulative_histogram` be more informative for this use case? (For finding a threshold.)

lgruen (approved):

scripts/validation/tob_wgs_call_rate_distribution.py
KatalinaBobowik: I tested 50 previously with the same number of variants, and it was a bit combed, so I dropped to 30. I can add in the `cumulative_histogram` as well. That's a good idea

KatalinaBobowik:
@lgruen Thanks for the suggestion on `is_defined` vs `is_missing`. I had a quick check and I get different results (not just the inverse) when using the two different options. So for example, on a variant that has 4 missing (NA) values, I get 16 out of 20, but 0 for `is_missing`. Is this normal behaviour?

lgruen:
> @lgruen Thanks for the suggestion on `is_defined` vs `is_missing`. I had a quick check and I get different results (not just the inverse) when using the two different options. So for example, on a variant that has 4 missing (NA) values, I get 16 out of 20, but 0 for `is_missing`. Is this normal behaviour?

That is odd, as the source code defines them using the same underlying `IsNA` function, but one is simply negated. I.e. `is_defined(x)` should be identical to `~is_missing(x)`.

```
def is_defined(expression) -> BooleanExpression:
    return ~apply_expr(lambda x: ir.IsNA(x), tbool, expression)

def is_missing(expression) -> BooleanExpression:
    return apply_expr(lambda x: ir.IsNA(x), tbool, expression)
```

Do you have an example that is easy to reproduce or maybe even just the output of that row that you're running this on, with the exact expression that you're using to get the 0 result for `is_missing`?

KatalinaBobowik:
Yeah, I was testing on an old, densified file (which we can get rid of if need be):
```
mt = hl.read_matrix_table('gs://cpg-tob-wgs-test/kat/v0/tob_wgs_densified_filtered.mt/')
example_locus = hl.parse_locus(‘chr1:54714’, ‘GRCh38’)
example_locus = mt.filter_rows(mt.locus == example_locus, keep = True)
# show 4 missing genotype calls
example_locus.GT.show(n_cols=20)
# add in defined values
example_locus = example_locus.annotate_rows(is_defined=hl.agg.sum(hl.is_defined(example_locus.GT)))
# add in missing values
example_locus = example_locus.annotate_rows(is_missing=hl.agg.sum(hl.is_missing(example_locus.GT)))

example_locus.is_missing.show()
# returns 0
example_locus.is_defined.show()
# returns 16
```

lgruen:
> Yeah, I was testing on an old, densified file (which we can get rid of if need be):
> 
> ```py
> # add in defined values
> example_locus = example_locus.annotate_rows(is_defined=hl.agg.sum(hl.is_defined(example_locus.GT)))
> # add in missing values
> example_locus = example_locus.annotate_rows(is_missing=hl.agg.sum(hl.is_missing(example_locus.GT)))
>
> example_locus.is_missing.show()
> # returns 0
> example_locus.is_defined.show()
> # returns 16
> ```

Thanks a lot for this awesome example! This is a bit of an unintuitive behavior of Hail, but the aggregation won't loop over missing values after they've been filtered. You can get around this by calling `unfilter_entries()` first:

```py
example_locus = example_locus.unfilter_entries()
example_locus = example_locus.annotate_rows(is_missing=hl.agg.sum(hl.is_missing(example_locus.GT)))
example_locus.is_missing.show()
# returns 4
```

KatalinaBobowik:
Oh, that makes sense and is super handy to know. Thank you! When you say that aggregation won't loop over missing values after they've been filtered, does that mean I need to change the ordering of assigning the call rate, or use `unfilter_entries()` in the plotting script?

scripts/validation/tob_wgs_call_rate_distribution.py
```diff
@@ -0,0 +1,38 @@
...
+    # filter out variants that didn't pass the VQSR filter
+    mt = mt.filter_rows(hl.len(hl.or_else(mt.filters, hl.empty_set(hl.tstr))) == 0)
+    # Calculate call rate and add to mt
+    mt = mt.annotate_rows(call_rate=hl.agg.sum(hl.is_defined(mt.GT)) / n_samples)
+    # plot distribution
+    p1 = hl.plot.histogram(mt.call_rate, range=(0, 1), bins=30, legend='Call Rate', title='Call Rate Distribution')
+    p1_filename = output_path('histogram_call_rate.png', 'web')
+    with hl.hadoop_open(p1_filename, 'wb') as f:
+        get_screenshot_as_png(p1).save(f, format='PNG')
+    p2 = hl.plot.cumulative_histogram(mt.call_rate, range=(0, 1), bins=30, legend='Call Rate', title='Call Rate Distribution')
```
lgruen, line 31: Note that this will evaluate all operations up to `mt.call_rate` _again_. I.e. it would be beneficial to `checkpoint` after `annotate_rows`.

lgruen:
> Oh, that makes sense and is super handy to know. Thank you! When you say that aggregation won't loop over missing values after they've been filtered, does that mean I need to change the ordering of assigning the call rate, or use `unfilter_entries()` in the plotting script?

I think your script is fine as is, since you're using `is_defined`, which works as expected.

While you're also filtering in your script, that's unrelated to *entries*. Those are treated in a special way to allow sparsity, which is explained in a bit more detail here: https://hail.is/docs/0.2/hail.MatrixTable.html#hail.MatrixTable.filter_entries

PR #138, opened 2022-09-30
Add suggestions as made by Daniel E and increase call rate

KatalinaBobowik:
Add suggestions as made by Daniel E in [this thread](https://centrepopgen.slack.com/archives/C01FBL6NWFN/p1664429696954009?thread_ts=1664149768.063269&cid=C01FBL6NWFN) and increase genotype call rate threshold, as per output from the [tob_wgs_call_rate_distribution.py](https://github.com/populationgenomics/tob-wgs/blob/main/scripts/validation/tob_wgs_call_rate_distribution.py) validation script (https://github.com/populationgenomics/tob-wgs/blob/main/scripts/validation/tob_wgs_call_rate_distribution.py) (see histogram [here](https://main-web.populationgenomics.org.au/tob-wgs/validation/tob_call_rate/v0/histogram_call_rate.png)).

### DIFF
--- a/scripts/eqtl_hail_batch/launch_eqtl_spearman.py
+++ b/scripts/eqtl_hail_batch/launch_eqtl_spearman.py
@@ -785,8 +785,8 @@ def run_spearman_correlation_scatter(
     spearman_df = t.to_pandas()
     spearman_df['round'] = '1'
     # change bp and round into int values
-    spearman_df['round'].astype(str).astype(int)
-    spearman_df['bp'].astype(str).astype(int)
+    spearman_df['round'].astype(int)
+    spearman_df['bp'].astype(int)
     # add celltype id
     celltype_id = cell_type.lower()
     spearman_df['cell_type_id'] = celltype_id
@@ -1285,8 +1285,8 @@ def run_scattered_conditional_analysis(
     celltype_id = cell_type.lower()
     adjusted_spearman_df['cell_type_id'] = celltype_id
     # change bp and round into int values
-    adjusted_spearman_df['round'].astype(str).astype(int)
-    adjusted_spearman_df['bp'].astype(str).astype(int)
+    adjusted_spearman_df['round'].astype(int)
+    adjusted_spearman_df['bp'].astype(int)
     # Correct for multiple testing using Storey qvalues
     # qvalues are used instead of BH/other correction methods, as they do not assume independence (e.g., high LD)
     pvalues = adjusted_spearman_df['p_value']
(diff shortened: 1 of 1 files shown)


scripts/eqtl_hail_batch/launch_eqtl_spearman.py
```diff
@@ -783,6 +784,9 @@ def spearman_correlation(df):
     # for front-end analysis
     spearman_df = t.to_pandas()
     spearman_df['round'] = '1'
+    # change bp and round into int values
+    spearman_df['round'].astype(str).astype(int)
+    spearman_df['bp'].astype(str).astype(int)
```
lgruen: Do you need two `astype` calls here (and below) or would `spearman_df['round'].astype(int)` work too?

scripts/eqtl_hail_batch/launch_eqtl_spearman.py
```diff
@@ -1268,12 +1272,21 @@ def spearman_correlation(df):
     # turn into hail table and annotate with global bp and allele info
     t = hl.Table.from_pandas(adjusted_spearman_df)
     t = t.annotate(global_bp=hl.locus(t.chrom, hl.int32(t.bp)).global_position())
+    # add in vep annotation
```
lgruen, line 1275: Was this change intentional? I'm asking because in the thread you said:

> For point 2, the functional annotation is there, it’s just not available for every variant

Does this change the annotation logic?

lgruen (changes requested):

scripts/eqtl_hail_batch/launch_eqtl_spearman.py
KatalinaBobowik: I think you're right, it would work. I didn't take notes for this, unfortunately, but I remember some part breaking when I had `round` as an int, so I was just being precautious and doing it after since it didn't fail locally. Should be the same though (and I'll change this).

KatalinaBobowik: I think you're right! I'll remove the redundant `astype(str)` bit

scripts/eqtl_hail_batch/launch_eqtl_spearman.py
KatalinaBobowik: I'm just addressing @daniaki 's point (slack message [here](https://centrepopgen.slack.com/archives/C01FBL6NWFN/p1664429696954009?thread_ts=1664149768.063269&cid=C01FBL6NWFN)) about making the subsequent rounds of the eQTL tables the same:

> Would it be simple to modify the round*/sigsnps/*.parquet outputs by adding functional_annotation...

So here I'm just adding in functional annotation for rounds 2-5

lgruen: Ah, thanks for explaining!

lgruen (approved):

PR #147, opened 2022-11-08
Count variants

annacuomo:


scripts/rv_expression_association/count_variants.py
```diff
@@ -0,0 +1,35 @@
...
+
+# This script aims to count the total number of variants (non-ref)
+# from the whole TOB-WGS dataset that are biallelic SNVs
+# and rare (MAF < 5%)
+
+import hail as hl
+from cpg_utils.hail_batch import dataset_path, init_batch
+
+# full TOB-WGS object
+MT = dataset_path('mt/v7.mt')
```
MattWellie: Instead of a constant value, could you configure this to take an input path using click, if it's likely that this is relevant for a different MT in future?

annacuomo: i tried

scripts/rv_expression_association/count_variants.py
```diff
@@ -0,0 +1,41 @@
...
+    )
+
+    # filter rare variants only (MAF < 5%)
+    mt = hl.variant_qc(mt)
+    mt = mt.filter_rows((mt.variant_qc.AF[1] < 0.05) & (mt.variant_qc.AF[1] > 0))
+
+    print(mt.count())
+
+
+if __name__ == '__count_variants__':
```
lgruen:
```suggestion
if __name__ == '__main__':
```

scripts/rv_expression_association/count_variants.py
```diff
@@ -0,0 +1,41 @@
...
+
+@click.command()
+@click.option('--mt-path', required=True)  # 'mt/v7.mt'
+def count_variants(
+    mt_path: str,
+):  
+    # read hail matrix table object (WGS data)
+    init_batch()
+    mt = hl.read_matrix_table(dataset_path(mt_path))
+    # densify
```
lgruen:
```suggestion
```

lgruen (approved):

PR #151, opened 2022-11-23
Copy geneloc files 7 and 22

KatalinaBobowik:
Copy geneloc file 7 to run for test gene CNTNAP2, as well as chr22 geneloc file for reference

### DIFF
--- a/scripts/data_transfer/copy_geneloc_genotype_files.sh
+++ b/scripts/data_transfer/copy_geneloc_genotype_files.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -ex
+
+# copy chromosome 7 and 22 geneloc files fmor main to test
+gsutil cp gs://cpg-tob-wgs-main/scrna-seq/grch38_association_files/gene_location_files/GRCh38_geneloc_chr7.tsv gs://cpg-tob-wgs-test/scrna-seq/grch38_association_files/gene_location_files/
+gsutil cp gs://cpg-tob-wgs-main/scrna-seq/grch38_association_files/gene_location_files/GRCh38_geneloc_chr22.tsv gs://cpg-tob-wgs-test/scrna-seq/grch38_association_files/gene_location_files/
+
+# copy genotype output from eQTL association pipeline into analysis to explore data
+gs://cpg-tob-wgs-main-tmp/scrna-seq/eqtl_output/v3-scale-test/genotype_table.mt gs://cpg-tob-wgs-main-tmp/scrna-seq/eqtl_output/v3-scale-test/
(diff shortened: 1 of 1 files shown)


scripts/data_transfer/copy_geneloc_genotype_files.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -ex
+
+# copy chromosome 7 and 22 geneloc files fmor main to test
+gsutil cp gs://cpg-tob-wgs-main/scrna-seq/grch38_association_files/gene_location_files/GRCh38_geneloc_chr7.tsv gs://cpg-tob-wgs-test/scrna-seq/grch38_association_files/gene_location_files/
+gsutil cp gs://cpg-tob-wgs-main/scrna-seq/grch38_association_files/gene_location_files/GRCh38_geneloc_chr22.tsv gs://cpg-tob-wgs-test/scrna-seq/grch38_association_files/gene_location_files/
+
+# copy genotype output from eQTL association pipeline into analysis to explore data
+gs://cpg-tob-wgs-main-tmp/scrna-seq/eqtl_output/v3-scale-test/genotype_table.mt gs://cpg-tob-wgs-main-tmp/scrna-seq/eqtl_output/v3-scale-test/
```
lgruen: This is copying to `tmp` instead of `analysis`, but more importantly we shouldn't copy this as a one-off, but change the pipeline to write to that destination instead.

```suggestion
```

scripts/data_transfer/copy_geneloc_genotype_files.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -ex
+
+# copy chromosome 7 and 22 geneloc files fmor main to test
```
lgruen:
```suggestion
# copy chromosome 7 and 22 geneloc files from main to test
```

lgruen (changes requested):

scripts/data_transfer/copy_geneloc_genotype_files.sh
KatalinaBobowik: Sorry, that was a copy paste typo on my end. I agree we should change where this is output in the pipeline, but for now to debug, should we still copy this file from `tmp` to `analysis`? Or should we only change this in the association script and regenerate it by rerunning?

lgruen: I'd prefer regenerating it, otherwise it'll be unclear later on where this came from.

lgruen (approved):
Thanks! Now just needs a quick update of the PR description.

PR #157, opened 2023-01-17
Explore genes

annacuomo:
Edit: Can you please have a look at the last change? The filtering step wasn't working as expected, and I was writing out an empty object.
I am now just saving a different mt for each gene, which may be even cleaner for downstream exploration, and uses the semi_join_rows syntax which should work. Testing it now!

Thanks!

-------- Older post for context below:

I am in the process of exploring the results from my first association run (using the [Hail Batch workflow](https://github.com/populationgenomics/cellregmap-pipeline/blob/main/batch.py) developed together).

As part of the pipeline, i saved to `analysis` Hail Tables summarising the variants considered for a given gene.

Now for a small set of genes for which i find significant association (n=22), I want to have a look at how these variants' frequencies are distributed in the OneK1K samples, so I need to look at the full Hail Matrix Table object instead.

The subset_variants.py script aims to subset the entire TOB-WGS mt object to just the variants tested for the genes I am interested in, and save to the analysis bucket so I can explore.

The script I am sure does not quite use the right syntax, and there may be a better way to deal with this altogether, but I thought I'd make a start! 

Thanks


scripts/rv_expression_association/subset_variants.py
```diff
@@ -0,0 +1,44 @@
...
+    # read ht objects to extract variant names
+    variants = []
+    genes_of_interest = genes.split(' ')
+    for gene in genes_of_interest:
+        ht_object_filename = f'gs://cpg-tob-wgs-main-analysis/tob_wgs_rv/pseudobulk_rv_association/summary_hts/{gene}_rare_promoter_summary.ht'
+        ht = hl.read_table(ht_object_filename)
+        variants.append(ht.locus.collect())  # I am nearly sure the syntax is wrong here
+
+    # read hail matrix table object (WGS data)
+    init_batch()  # does this need to be above?
```
MattWellie: This will have to come before the first read_table call, else Hail might load an unexpected reference genome

scripts/rv_expression_association/subset_variants.py
```diff
@@ -0,0 +1,44 @@
...
+    output_mt_path: str,
+):
+
+    # read ht objects to extract variant names
+    variants = []
+    genes_of_interest = genes.split(' ')
+    for gene in genes_of_interest:
+        ht_object_filename = f'gs://cpg-tob-wgs-main-analysis/tob_wgs_rv/pseudobulk_rv_association/summary_hts/{gene}_rare_promoter_summary.ht'
+        ht = hl.read_table(ht_object_filename)
+        variants.append(ht.locus.collect())  # I am nearly sure the syntax is wrong here
```
MattWellie:
```suggestion
        variants.extend(ht.locus.collect())  # I am nearly sure the syntax is wrong here
```

`append` would add the whole list, so you'd have a list-of-lists. I think you want `extend` so you end up with one unified list for the `mt.locus in variants` test below?

scripts/rv_expression_association/subset_variants.py
```diff
@@ -0,0 +1,44 @@
...
+    mt = hl.read_matrix_table(dataset_path(input_mt_path))
+    mt = hl.experimental.densify(mt)
+
+    # filter out low quality variants and consider biallelic variants only (no multi-allelic, no ref-only)
+    mt = mt.filter_rows(mt.locus in variants)
+
+    # save mt
+    mt.write(
+        output_mt_path, overwrite=True
+    )  # how do I specify the analysis bucket again?
```
MattWellie: If you're using a full path as the input, just include `main/test-analysis` in the bucket path. If you're forming it using `output_path`, you'd need to add `analysis` as the suffix, e.g.

```python
Pathy_McPathface = output_path('significant_genes_burden_max.mt', 'analysis')
```

scripts/rv_expression_association/subset_variants.py
annacuomo: yea i thought that may be the case, great thanks!

scripts/rv_expression_association/subset_variants.py
annacuomo: awesome!

scripts/rv_expression_association/subset_variants.py
annacuomo: so I am doing the former i think, would the latter be preferable?

MattWellie: it's up to you 🤷‍♂️ 

`output_path` means that you can change the `-o` analysis-runner argument and it will automatically direct all the output generated to the defined folder structure. With the current solution you'll have to manually change that path for the same result... There's nothing better or worse between the two options really!

scripts/rv_expression_association/subset_variants.py
```diff
@@ -0,0 +1,44 @@
...
+def subset_variants(
+    input_mt_path: str,
+    genes: str,
+    output_mt_path: str,
+):
+    init_batch()
+
+    # read ht objects to extract variant names
+    variants = []
+    genes_of_interest = genes.split(' ')
```
MattWellie: Have you tested this somewhere? I don't know how well click is going to handle whitespace-separated arguments once they're passed via analysis-runner

scripts/rv_expression_association/subset_variants.py
annacuomo: ok!

scripts/rv_expression_association/subset_variants.py
annacuomo: So I think this should work as it did for the hail batch workflow? (https://github.com/populationgenomics/cellregmap-pipeline/blob/main/batch.py#L704)

And no I tried to test but I don't have access to the main-analysis bucket from main I think! Is the best approach to use [your script](https://github.com/populationgenomics/analysis-runner/blob/main/scripts/subset_hail_table.py) to copy a couple of those summary hts to `test` so I can test this?

scripts/rv_expression_association/subset_variants.py
```diff
@@ -0,0 +1,44 @@
...
+    for gene in genes_of_interest:
+        ht_object_filename = f'gs://cpg-tob-wgs-test-analysis/tob_wgs_rv/pseudobulk_rv_association/summary_hts/{gene}_rare_promoter_summary.ht'
```
illusional: The base for this path should be derived, or passed in to support both main / test namespaces

scripts/rv_expression_association/subset_variants.py
illusional: \+1 to doing the latter here

scripts/rv_expression_association/subset_variants.py
```diff
@@ -0,0 +1,44 @@
...
+# pylint: disable=no-value-for-parameter,missing-function-docstring,missing-module-docstring
+
+import click
+import hail as hl
+from cpg_utils.hail_batch import dataset_path, init_batch
+
+
+@click.command()
+@click.option('--input-mt-path', required=True)  # 'mt/v7.mt'
+@click.option('--genes', required=True)  # 'LMNA'
```
illusional: Should say that they're space separated, provided as one argument.

scripts/rv_expression_association/subset_variants.py
MattWellie:
```suggestion
@click.command()
@click.option('--input-mt-path', required=True)  # 'mt/v7.mt'
@click.option(
    '--output-mt-path', required=True
)  # 'gs://cpg-tob-wgs-main-analysis/tob_wgs_rv/pseudobulk_rv_association/significant_genes_burden_max.mt'
@click.argument('genes_of_interest', nargs=-1)  # positional argument(s)
def subset_variants(
    input_mt_path: str,
    output_mt_path: str,
    genes_of_interest: tuple[str],
):
    """
    genes_of_interest as a positional argument, accepting
    one or more genes
    """
    
    init_batch()

    # read ht objects to extract variant names
    variants = []
```

syntax for taking genes as a list of strings instead of one string and splitting manually

annacuomo: i dont know i seem to remember we tried this previously and failed?

MattWellie: This is the difference between `click.option` and `click.argument` 

- options can take multiple inputs, but due to how they're parsed they have to be a fixed/known number. 

- arguments are positional, and you can take an arbitrary number of input values (nargs=-1) - these are collected as a tuple

MattWellie: I made a little test script to try this out locally:

```
import click


@click.argument('g', nargs=-1)
@click.command()
def main(g):
    print(g)

if __name__ == '__main__':
    main()
```

> python click_test.py a a a
('a', 'a', 'a')

scripts/rv_expression_association/subset_variants.py
```diff
@@ -0,0 +1,48 @@
...
+
+import click
+import hail as hl
+from cpg_utils.hail_batch import dataset_path, init_batch, output_path
+
+
+@click.command()
+@click.option('--input-mt-path', required=True)  # 'mt/v7.mt'
+@click.option('--genes', required=True, help='List of genes to consider. Space separated, as one argument.')  # 'LMNA'
+@click.option('--output=mt-name', required=True)  # 'significant_genes_burden_max.mt'
```
illusional: Is this supposed to have the equals sign?

illusional (approved):
One MUST action comment about parma name, but apart from that looks good to me!

scripts/rv_expression_association/subset_variants.py
annacuomo: nope that's embarrassing and should be fixed now lol

MattWellie (approved):

MattWellie (approved):
quite a few print statements you may not want on main, but the logic looks sound

annacuomo:
> quite a few print statements you may not want on main, but the logic looks sound

yes i will remove those when I am done testing and satisfied with the output, thanks!

PR #158, opened 2023-02-14
Create copy_atac_annos.sh

annacuomo:
Moving chromatin annotations from `main-upload` to `main`, subset of chr 21 only to `test`.

### DIFF
--- a/scripts/data_transfer/copy_atac_annos.sh
+++ b/scripts/data_transfer/copy_atac_annos.sh
@@ -3,5 +3,7 @@
 set -ex
 
 gsutil -m mv gs://cpg-tob-wgs-main-upload/open_chromatin_annotation/predicted_l1_celltypes_avg_peaks.csv gs://cpg-tob-wgs-main/tob_wgs_rv/open_chromatin_annotation/
+gsutil -m mv gs://cpg-tob-wgs-main-upload/open_chromatin_annotation/predicted_l2_celltypes_avg_peaks.csv gs://cpg-tob-wgs-main/tob_wgs_rv/open_chromatin_annotation/
+gsutil -m mv gs://cpg-tob-wgs-main-upload/open_chromatin_annotation/predicted_l3_celltypes_avg_peaks.csv gs://cpg-tob-wgs-main/tob_wgs_rv/open_chromatin_annotation/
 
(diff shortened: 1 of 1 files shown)


scripts/data_transfer/copy_atac_annos.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -ex
+
+gsutil -m mv gs://cpg-tob-wgs-main-upload/open_chromatin_annotation/predicted_l1_celltypes_avg_peaks.csv gs://cpg-tob-wgs-main/tob_wgs_rv/open_chromatin_annotation/
+gsutil -m mv gs://cpg-tob-wgs-main-upload/open_chromatin_annotation/predicted_l2_celltypes_avg_peaks.csv gs://cpg-tob-wgs-main/tob_wgs_rv/open_chromatin_annotation/
+gsutil -m mv gs://cpg-tob-wgs-main-upload/open_chromatin_annotation/predicted_l3_celltypes_avg_peaks.csv gs://cpg-tob-wgs-main/tob_wgs_rv/open_chromatin_annotation/
```
lgruen:
```suggestion
gsutil -m mv "gs://cpg-tob-wgs-main-upload/open_chromatin_annotation/predicted_l*_celltypes_avg_peaks.csv" gs://cpg-tob-wgs-main/tob_wgs_rv/open_chromatin_annotation/
```

scripts/data_transfer/copy_atac_annos.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -ex
+
+gsutil -m mv gs://cpg-tob-wgs-main-upload/open_chromatin_annotation/predicted_l1_celltypes_avg_peaks.csv gs://cpg-tob-wgs-main/tob_wgs_rv/open_chromatin_annotation/
```
lgruen: What does `tob_wgs_rv` stand for? Should we omit that?

lgruen (approved):

scripts/data_transfer/copy_atac_annos.sh
annacuomo: TOB whole-genome sequencing, rare variant (analysis) - yes I am still fairly unsure re: data paths. For example, this is in principle good for any scRNA-seq from blood, which we'll have for TOB, BioHEART and many others.. but this is the TOB repo anyways, so for now the application when I'll use these data is to prioritise rare variants (from WGS) to test for associations

PR #159, opened 2023-02-14
script for subsetting requested CRAM files

MattWellie:
as described

relevant BED files in `gs://cpg-tob-wgs-test/subset_beds`

### DIFF
--- a/scripts/subset_crams.py
+++ b/scripts/subset_crams.py
@@ -63,7 +63,7 @@ def main(beds: str):
     cram_map = get_cram_files(list(sample_map.values()))
 
     # set an output path to write files to
-    release_cram = 'gs://cpg-tob-wgs-release/cram'
+    release_cram = 'gs://cpg-tob-wgs-release/MRR_cram_extracts/2023_01_30'
 
     # set the image and reference to use
     samtools_image = get_config()['images']['samtools']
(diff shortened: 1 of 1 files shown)


scripts/subset_crams.py
```diff
@@ -0,0 +1,121 @@
...
+    ext_ids = {file.name.split('.', maxsplit=1)[0]: str(file) for file in bed_files}
+
+    # get lookup of all CPG: Ext IDs for these samples
+    sample_map = get_samples_from_metamist(list(ext_ids.keys()))
+
+    # get all (latest) CRAM files for these samples
+    cram_map = get_cram_files(list(sample_map.values()))
+
+    # set an output path to write files to
+    release_cram = 'gs://cpg-tob-wgs-release/cram'
```
cassimons: Can we make the out dir "mini-cram" or similar?

cassimons (approved):
Looks good to me.  I have no idea if this is where these types of scripts should live? 

In terms of output name, we should probably add some pre-suffix to the crams indicating their purpose: "{extid}.ror.cram" or "{extid}.mini.cram"

scripts/subset_crams.py
MattWellie: Yup, I was wondering if we had an identifiable event/purpose for this request? Or we could add the date to this output folder, then incorporate 'mini' into the individual file name

cassimons: a date based output might make sense as they are specific regions for manual inspection by the RoR team.

PR #162, opened 2023-02-16
Open chromatin annotate

annacuomo:
Script to annotate variants based on open chromatin info from snATAC-seq data

The open-ness of chromatin within a genomic-region (peak) is measured as the average read count for a specific cell type from single-nucleus ATAC-seq data from PBMCs from equivalent samples to the TOB cohort.

Successfully ran in `test` now ([log](https://batch.hail.populationgenomics.org.au/batches/421939/jobs/1)), can you check and then merge?

p.s. the lint errors are coming from subset_crams.py, launch_eqtl_spearman.py, peer.py from what I can tell?


MattWellie (approved):
LGTM

PR #164, opened 2023-02-16
update to use GCS files without copying

MattWellie:
Update to use direct access of GCS CRAMs rather than copy-in

Prior to merging, intent is to tidy up some linting errors

### DIFF
--- a/scripts/subset_crams.py
+++ b/scripts/subset_crams.py
@@ -6,6 +6,8 @@ Righto then, what do?
 - for each sample, create a job to create a mini-CRAM using its relevant BED
 - write the resulting files to the main bucket
 
+Uses GCS_OAUTH_TOKEN within the samtools container to read GCS CRAMs directly
+
 Requires manual transfer later to the release bucket
 """
 
@@ -14,6 +16,7 @@ import click
 
 from cpg_utils import to_path
 from cpg_utils.config import get_config
+from cpg_utils.hail_batch import authenticate_cloud_credentials_in_job
 from cpg_workflows.batch import get_batch, dataset_path
 
 from sample_metadata.model.analysis_type import AnalysisType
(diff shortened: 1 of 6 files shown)


scripts/subset_crams.py
```diff
@@ -103,13 +107,15 @@ def main(beds: str):
         # -C: output CRAM
         # -L: target regions file, specific to sample
         # --write-index: ...
+        # sets GCS_OAUTH_TOKEN to access data directly
         cram_job.command(
```
lgruen, line 114: Probably doesn't make a big difference, but would specifying `@` here allow for some multi-threading? (Assuming the 16G of RAM will result in multiple cores being available.

scripts/subset_crams.py
```diff
@@ -103,13 +107,15 @@ def main(beds: str):
...
+            'GCS_OAUTH_TOKEN=$(gcloud auth application-default print-access-token) '
             'samtools view '
             f'-T {batch_reference} '
             f'-L {sample_bed} '
             '-C --write-index '
             f'-o {cram_job.output_cram["cram"]} '
```
lgruen: Could probably also write directly?

scripts/subset_crams.py
```diff
@@ -80,15 +83,16 @@ def main(beds: str):
...
         cram_job = batch.new_job(name=f'subset CRAM {ext_id}/{cpg_id}')
+
+        # authenticate credentials in job
+        authenticate_cloud_credentials_in_job(cram_job)
+
+        # run the job inside the samtools image
         cram_job.image(samtools_image)
 
         # big storage, probably not huge on compute (small regions)
         cram_job.storage('60G')
```
lgruen: This really just needs to be enough to fit the reference, right?

lgruen (approved):

scripts/subset_crams.py
MattWellie: Reference and any temp/intermediate files the threading would cause - I'll reduce to much lower than this

PR #173, opened 2024-01-25
Update linting versions, add ruff + cpg-id-checker

illusional:
This repo is a little stale, needs a bit of TLC :)


MattWellie (approved):
I've added an extra commit for completion - deleting the `.flake8` config file from the project root, and replacing all the dangling `# flake8: noqa: ...` which ruff will recognise, with an explicit `# ruff: noqa: ...`. 

Can you double check and see if you're happy with it

PR #174, opened 2024-04-18
Enrich tob/bioheart metadata with sequencing dates

amyminiter:
This PR includes a script that enriches existing tob/bioheart records in metamist with sequencing_dates in the meta field. 

See: https://centrepopgen.slack.com/archives/C018KFBCR1C/p1712796597218319

All assays are retrieved from metamist associated with the tob-wgs/bioheart datasets. These are then referenced with tube IDs provided in two separate .csv files. Assays associated with tubes that are referenced in the .csv files are updated with the associated sequencing date in the .csv file.

### DIFF
--- a/scripts/metadata_enrichment/tob_bioheart_sequencing_dates.py
+++ b/scripts/metadata_enrichment/tob_bioheart_sequencing_dates.py
@@ -188,6 +188,10 @@ def compare_tubes_metamist_excel(
     fluidx_to_assay_ids: defaultdict,
     fluidx_to_sequencing_date: dict,
 ):
+    """
+    Includes diagnostic code used to identify tubes missing between metamist and the
+    provided sequencing date manifests. Prints output to console.
+    """
     # Create a set from the fluidX identifiers only that were extracted from metamist
     metamist_set_fluidx = set(fluidx_to_assay_ids.keys())
 
--- a/scripts/metadata_enrichment/tob_bioheart_sequencing_dates.py
+++ b/scripts/metadata_enrichment/tob_bioheart_sequencing_dates.py
@@ -27,6 +27,14 @@ QUERY_PROJECT_ASSAYS = gql(
 
 # TODO: Create separate function for default dict creation
 def query_metamist():
+    """
+    Query metamist for bioheart and tob-wgs datasets and map all tube ids to samples.
+    Later calls append_dictionaries(), which collates dicts from the two projects.
+
+    :return: Dictionary of lists, mapping tube IDs to samples retrieved from metamist
+            for both bioheart and tob-wgs datasets
+    :rtype: defaultdict
+    """
     # gives us json output from graphql query
     # Results are seen in graphql interface as a dictionary
     query_result_tob = query(QUERY_PROJECT_ASSAYS, variables={'datasetName': 'tob-wgs'})
@@ -39,37 +47,37 @@ def query_metamist():
     tob_samples = query_result_tob['project']['samples']
     bioheart_samples = query_result_bioheart['project']['samples']
 
-    # TOB-WGS
-    # Create list of all assays from tob_samples (at an individual level)
+    # Create default dicts for the two datasets: maps tube IDs to samples
+    tob_kccg_id = 'KCCG FluidX tube ID'
+    fluidx_to_assay_ids_tob = create_default_dict(tob_samples, tob_kccg_id)
+    print(f'Type for samples is: {type(tob_samples)} OR {type(bioheart_samples)}')
+    bioheart_kccg_id = 'fluid_x_tube_id'
+    fluidx_to_assay_ids_bioheart = create_default_dict(
+        bioheart_samples,
+        bioheart_kccg_id,
+    )
+
+    return append_dictionaries(fluidx_to_assay_ids_tob, fluidx_to_assay_ids_bioheart)
+
+
+def create_default_dict(samples: list, kccg_id: str) -> defaultdict:
     assays = []
-    for sample in tob_samples:
+    for sample in samples:
         assays.extend(sample['assays'])
 
     # Use default dict to group assays with fluid X tube metadata
-    fluidx_to_assay_ids_tob = defaultdict(list)
+    tube_to_assays_defaultdict = defaultdict(list)
     for assay in assays:
         # per assay, extract assay ID and fluid X tubeID
         # fluid tube id is the key; list contains assay IDs
-        fluidx_tube_id = assay['meta'].get('KCCG FluidX tube ID')
-        # Check that FluidX tube is not null TODO
-        fluidx_to_assay_ids_tob[fluidx_tube_id].append(assay['id'])
-
-    # BIOHEART
-    assays_bioheart = []
-    for sample in bioheart_samples:
-        assays_bioheart.extend(sample.get('assays'))
-
-    # Use default dict to group assays with fluid X tube metadata
-    fluidx_to_assay_ids_bioheart = defaultdict(list)
-    for assay in assays_bioheart:
-        # per assay, extract assay ID and fluid X tubeID
-        # fluid tube id is the key; list contains assay IDs
-        fluidx_tube_id_bioheart = assay.get('meta').get('fluid_x_tube_id')
-        fluidx_to_assay_ids_bioheart[fluidx_tube_id_bioheart.split('_')[1]].append(
-            assay.get('id'),
-        )
-
-    return append_dictionaries(fluidx_to_assay_ids_tob, fluidx_to_assay_ids_bioheart)
+        fluidx_tube_id = assay['meta'].get(kccg_id)
+        if kccg_id.endswith('tube_id'):
+            tube_to_assays_defaultdict[fluidx_tube_id.split('_')[1]].append(
+                assay.get('id'),
+            )
+        else:
+            tube_to_assays_defaultdict[fluidx_tube_id].append(assay['id'])
+    return tube_to_assays_defaultdict
 
 
 def append_dictionaries(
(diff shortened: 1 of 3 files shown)


amyminiter:
Script yet to be trialled on test environment. I am currently creating a test environment locally.

scripts/metadata_enrichment/tob_bioheart_sequencing_dates.py
```diff
@@ -0,0 +1,234 @@
...
+    bioheart_kccg_id = 'fluid_x_tube_id'
+    fluidx_to_assay_ids_bioheart = create_default_dict(
+        bioheart_samples,
+        bioheart_kccg_id,
+    )
+
+    return append_dictionaries(fluidx_to_assay_ids_tob, fluidx_to_assay_ids_bioheart)
+
+
+def create_default_dict(samples: list, kccg_id: str) -> defaultdict:
```
EddieLF: I would rename this function something like "`create_fluidx_id_to_assay_ids_dict`" - or something to that effect. Current name is a bit too broad imo.

scripts/metadata_enrichment/tob_bioheart_sequencing_dates.py
```diff
@@ -0,0 +1,234 @@
...
+    """
+    Creates defaultdict mapping kccg tube ids to all related samples.
```
EddieLF: This docstring should say that the function maps the tube ID to all related assays, not samples. Same with the `:return:` below.

scripts/metadata_enrichment/tob_bioheart_sequencing_dates.py
```diff
@@ -0,0 +1,234 @@
...
+    set_tob = set(fluidx_to_assay_ids_tob.keys())
+    set_bioheart = set(fluidx_to_assay_ids_bioheart.keys())
+
+    if len(set_tob.intersection(set_bioheart)) == 0:
+        # Append bioheart dict to tob dict
+        for d in dicts_to_merge:
+            for key, value in d.items():
+                result[key].append(value)
+        return result
+    else:  # noqa: RET505
```
EddieLF: Since the `if` condition will always lead to a `return`, there is no need for an `else` here (hence the RET505 flake8 message).

Personally I think that it looks clearer with the `else` so feel free to leave it as is, but just in case it wasn't clear why the RET505 flag is appearing, it's because the else is considered "superfluous".

scripts/metadata_enrichment/tob_bioheart_sequencing_dates.py
```diff
@@ -0,0 +1,234 @@
...
+    # construct API update calls
+    # Iterate through the fluidX_to_assay_IDs dictionary because this is representative of what's already in metamist
+    # That is: fluidX_to_assay_IDs groups assays by fluidX_tube_id
+
+    assay_api = AssayApi()
+    assays_to_update = []
+
+    for fluidx_id, assay_ids in fluidx_to_assay_ids.items():
+        # Get sequencing date for each fluidX id from fluidX_to_sequencing_date dict
+        if fluidx_id:
```
EddieLF: You can reduce a level of indentation here with a so-called "early return", which is using a continue to go to the next stage of the loop if a condition is not met. This removes the need for an if-else and knocks out one indentation of nesting. 

Something like:
```python
for fluidx_id, assay_ids in fluidx_to_assay_ids.items():
    if not fluidx_id:
        print(f'*****\nNO TUBE ID FOR : {fluidx_id} and assays {assay_ids}')
        continue
    try:
        ...

File: scripts/metadata_enrichment/tob_bioheart_sequencing_dates.py [hunk 5]
```diff
@@ -0,0 +1,234 @@
...
+                # Create object required to update assay
+                date = fluidx_to_sequencing_date[fluidx_id]
+            except KeyError:
+                print(f'Tube ID {fluidx_id} is not in the sequencing manifest')
+            else:
+                for assay_id in assay_ids:
+                    updated_assay = AssayUpsert()
+                    updated_assay['id'] = assay_id
+                    updated_assay['meta'] = {}
+                    updated_assay['meta']['fluid_x_tube_sequencing_date'] = date
```
EddieLF:
```suggestion
                    updated_assay = AssayUpsert(
                        id = assay_id,
                        meta = {'fluid_x_tube_sequencing_date': date}
                    )
```

scripts/metadata_enrichment/tob_bioheart_sequencing_dates.py
```diff
@@ -0,0 +1,234 @@
...
+                    assays_to_update.append(updated_assay)
+                    print(f'API call added for {assay_id}')
+
+        else:
+            print(f'*****\nNO TUBE ID FOR : {fluidx_id} and assays {assay_ids}')
+
+    return await assay_api.upsert_assay_async(assay_upsert=list(assays_to_update))
```
EddieLF: I dont think this way of calling the API endpoint will work. This endpoint is expecting a single AssayUpsert object as input, so it would need to go inside the for loop above. This is definitely one that we would want to `gather` a number of individual async calls. Let's chat about it tomorrow.

EddieLF (commented):
Hey Amy, a few nitpicky and notes here but it mostly looks good! However I suspect the upsert_assay_async API call will need changing before this works. If you have some time in the next few days let's book in a call and go through this. Cheers

scripts/metadata_enrichment/tob_bioheart_sequencing_dates.py
amyminiter: Very good idea. Updated.

scripts/metadata_enrichment/tob_bioheart_sequencing_dates.py
amyminiter: Fixed!

scripts/metadata_enrichment/tob_bioheart_sequencing_dates.py
amyminiter: I still want the else block. Suspect this has arisen from some lazy programming. Changed it to return at the end of the function instead.

scripts/metadata_enrichment/tob_bioheart_sequencing_dates.py
amyminiter: That is a _lot_ cleaner

scripts/metadata_enrichment/tob_bioheart_sequencing_dates.py
amyminiter: Implemented. Great suggestion.

scripts/metadata_enrichment/tob_bioheart_sequencing_dates.py
```diff
@@ -0,0 +1,247 @@
...
+    else:
+        print(
+            f'Error: these keys intersect: {set_tob.intersection(set_bioheart)}',
+        )
+        result = defaultdict()
+
+    return result
+
+
+# TODO: Update sequencing date with correct value!
```
EddieLF: Is this still todo?

scripts/metadata_enrichment/tob_bioheart_sequencing_dates.py
```diff
@@ -0,0 +1,247 @@
...
+    )
+
+    # Insert sequencing date value into new dictionary fluidX_to_sequencing_date. Key is FluidX ID
+    return pd.Series(
+        aggregated_df['Date (YY/MM/DD)'].values,
+        index=aggregated_df.fluidX_id,
+    ).to_dict()
+
+
+# TODO: combine api calls and
```
EddieLF: Same here

scripts/metadata_enrichment/tob_bioheart_sequencing_dates.py
```diff
@@ -0,0 +1,247 @@
...
+    assays_no_tubes = fluidx_to_assay_ids.get(None, 'empty')
+    print(assays_no_tubes)
+    print(len(assays_no_tubes[0]))
+
+
+async def main():
+    fluidx_to_assay_ids = query_metamist()
+    if len(fluidx_to_assay_ids) != 0:
+        fluidx_to_sequencing_date = extract_excel()
+        await upsert_sequencing_dates(
```
EddieLF: I think you could add a commented out line or two here like:
```python
# Uncomment this to compare the data in metamist to the manifests
# compare_tubes_metamist_excel(fluidx_to_assay_ids, fluidx_to_sequencing_date)
```

Because currently the compare_tubes_metamist_excel function is unused, which is fine, the docstring explains it's there for debugging, but it might be good to show where in the code it would be used.

You could go even further and add a click option like `@click.option('--debug', '-d', is_flag=True)`
And then in the `main` function:
```python
if debug:
   compare_tubes_metamist_excel(fluidx_to_assay_ids, fluidx_to_sequencing_date)
```

Refactoring and adding click isn't necessary, but still I think it would be good to include the function somewhere, even if it's commented out for the purposes of running the script on production data.

EddieLF (approved):
Hey Amy, this looks good! Approving, noting I've left a few comments about some inline #TODOs that could be cleaned up before merging, and also a note on the unused function at the bottom of the script. Nice job :)

scripts/metadata_enrichment/tob_bioheart_sequencing_dates.py
amyminiter: It is not! Comment deleted

scripts/metadata_enrichment/tob_bioheart_sequencing_dates.py
amyminiter: Have added the two comments as suggested.

I agree that `click` makes this more reusable in the longer term but I'm going to leave this as-is for now.

amyminiter: I changed my mind and added the click module 👯

PR #175, opened 2024-05-23
Update logic in append_dictionaries()

amyminiter:
Bug fix. Updated aggregation of two DefaultDicts in the function append_dictionaries(). 

Using append() added an unnecessary extra layer of lists, which later makes the assay_id unreadable in a nested list. Updating extend avoids this additional layer.


EddieLF (approved):
LGTM

PR #177, opened 2024-05-28
Separate datasets tob bioheart enrichment

amyminiter:
Following the permissions issue last week where I tried to run script with the `bioheart` dataset on `tob-wgs` data, this PR updates the enrichment script for `tob-wgs`/`bioheart`. The script no longer upserts assays for both datasets simultaneously and will only update a single dataset at a time.

There are also some minor changes to a dummy enrichment script that I created to populate my local test environment. You can largely ignore these changes, or you may recommend that these be deleted from the PR.


EddieLF (approved):
Nice changes @amyminiter! I think making the script flexible enough to take either dataset as input and handle the upserts was the right choice here.

PR #186, opened 2024-08-29
Dragmap Parity: Fixing 'None' samples showing up

michael-harper:
Removing samples keyed as None where there are no common samples identified between nagim and new. These non-shared samples are converted to 'None' and when performing `hl.concordance()` these multiple 'None' values are seen as duplicates and messing up the calculation.

Also adding the `hl.split_multi_hts()` function to split the `nagim_mt` to ensure only bi-allelics are being compared


silkm (approved):
Note to ourselves for later: We assume here that `split_multi_hts` does not change any loci positions as the split does on a vds (and there we use filter_changed_loci).
If we get an unusually high number of mismatches, we could annotate the MatrixTable with the loci integer as a backup before splitting, and check if the loci position changes after the split. Likely not an issue though as the Hail help pages don't mention this would be a behaviour for split_multi_hts.

